PR #57, opened 2016-05-17
VS2013 - C# 6 auto-property initializers - issue

Johnnyxy:
Hi,
I tried to compile the current code with Visual Studio 2013 (Update 3).
The compilation fails because of the use of a C# 6 feature called "auto-property initializers" which is not supported by VS2013 (Update 3) which uses C# 5.0.

One possible fix is to use a constructor.

Regards

### DIFF
--- a/S7.Net.Common/Types/DataItem.cs
+++ b/S7.Net.Common/Types/DataItem.cs
@@ -6,8 +6,13 @@
         public VarType VarType { get; set; }
         public int DB { get; set; }
         public int StartByteAdr { get; set; }
-        public int Count { get; set; } = 1;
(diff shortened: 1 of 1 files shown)


killnine:
Thanks, I think it might be worth creating a 'code conventions' document or section in the Readme.md to protect against features like this. C#6 has some nice functionality but unless there's a big reason for using such features, creating a more accessible platform (compiling as 4.5.2) seems more important.

mesta1:
I setup an appveyor build with VS 2013 and framework 3.5 (see readme.md). This should be enough to check possible errors on pull requests.

Issue #76, opened 2017-05-23
Write to bit should be atomic


sam0737:
Currently bit writing like DB100.DBX200.3 is done by reading a byte from DB100.DBD200, modify and write the whole back. This is not atomic and unsafe. Any modification in between by other system or PLC itself could be overwritten.

The protocol supports BIT addressing and writing only one bit. (I am going to create a PR)

mesta1:
Thanks for the PR and I would really love to accept it, but it breaks a lot of code.
I will make a review and if you can adapt it I will push into the codebase.

PR #78, opened 2017-05-23
Support atomic bit writing protocol, instead of read-modify-write

sam0737:
PR for #76

### DIFF
--- a/S7.Net/PLC.cs
+++ b/S7.Net/PLC.cs
@@ -540,14 +540,19 @@ namespace S7.Net
         /// <param name="startByteAdr">Start byte address. If you want to write DB1.DBW200, this is 200.</param>
         /// <param name="value">Bytes to write. If more than 200, multiple requests will be made.</param>
         /// <returns>NoError if it was successful, or the error is specified</returns>
-        public ErrorCode WriteBytes(DataType dataType, int db, int startByteAdr, byte[] value)
+        public ErrorCode WriteBytes(DataType dataType, int db, bool bitMode, int startByteAdr, int bitAdr, byte[] value)
         {
             int localIndex = 0;
             int count = value.Length;
+            if (bitMode && value.Length != 1)
+            {
+                // The code does not handle multiple-bit writing yet
+                throw new Exception(string.Format("Addressing Error: Multi-bit writing is not supported yet"));
+            }
             while (count > 0)
             {
                 var maxToWrite = (int)Math.Min(count, 200);
-                ErrorCode lastError = WriteBytesWithASingleRequest(dataType, db, startByteAdr + localIndex, value.Skip(localIndex).Take(maxToWrite).ToArray());
+                ErrorCode lastError = WriteBytesWithASingleRequest(dataType, db, bitMode, startByteAdr + localIndex, bitAdr, value.Skip(localIndex).Take(maxToWrite).ToArray());
                 if (lastError != ErrorCode.NoError)
                 {
                     return lastError;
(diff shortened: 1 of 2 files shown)


S7.Net/PLC.cs
```diff
@@ -540,14 +540,19 @@ public void ReadClass(object sourceClass, int db, int startByteAdr = 0)
         /// <param name="startByteAdr">Start byte address. If you want to write DB1.DBW200, this is 200.</param>
         /// <param name="value">Bytes to write. If more than 200, multiple requests will be made.</param>
         /// <returns>NoError if it was successful, or the error is specified</returns>
-        public ErrorCode WriteBytes(DataType dataType, int db, int startByteAdr, byte[] value)
+        public ErrorCode WriteBytes(DataType dataType, int db, bool bitMode, int startByteAdr, int bitAdr, byte[] value)
```
mesta1: It is better to write a separate function for WriteBit, and keep WriteBytes as it is. 
WriteBytes writes bytes, WriteBit writes bit.
WriteBit(int db, int startByteAdr, int bitAdr, bool value)

S7.Net/PLC.cs
```diff
@@ -540,14 +540,19 @@ public void ReadClass(object sourceClass, int db, int startByteAdr = 0)
...
         {
             int localIndex = 0;
             int count = value.Length;
+            if (bitMode && value.Length != 1)
```
mesta1: With separate functions, this check ins not more needed, because value becomes a boolean.

S7.Net/PLC.cs
```diff
@@ -684,14 +694,8 @@ public ErrorCode Write(string variable, object value)
...
-                                }
-                                byte b = (byte)Read(DataType.DataBlock, mDB, mByte, VarType.Byte, 1);
-                                if (Convert.ToInt32(value) == 1)
-                                    b = (byte)(b | (byte)Math.Pow(2, mBit)); // Bit setzen
-                                else
-                                    b = (byte)(b & (b ^ (byte)Math.Pow(2, mBit))); // Bit rücksetzen
-
-                                return Write(DataType.DataBlock, mDB, mByte, (byte)b);
+                                }
+                                return Write(DataType.DataBlock, mDB, mByte, (bool)value, mBit);
```
mesta1: Need to check if the method before was working with true/false or with 0/1.

S7.Net/Types/Boolean.cs
```diff
@@ -8,31 +8,12 @@ namespace S7.Net.Types
     public static class Boolean
     {
         /// <summary>
-        /// Returns the value of a bit in a bit, given the address of the bit
+        /// Converts a bit to byte array
         /// </summary>
-        public static bool GetValue(byte value, int bit)
```
mesta1: These are all public methods. We can't just remove these, but we have to mark the old one as obsolete and add the new ones.

S7.Net/PLC.cs
sam0737: There are so many magic numbers for the TPKT/COTP/S7 protocol that I am not exactly comfortable in splitting it.

S7.Net/PLC.cs
sam0737: I actually don't know for sure how does writing multiple bits work.

S7.Net/Types/Boolean.cs
sam0737: Ya I can keep them.

S7.Net/PLC.cs
mesta1: This is the public WriteBytes, which doesn't contain any code of the protocol. 
The method with magic numbers is WriteBytesWithASingleRequest, which is fine.

mesta1:
Let me know if you have time to update the PR, I would really like to merge these changes into the codebase.

sam0737:
Sorry I am on something else so might not be available to make these changes recently.

uknickwilliams:
Is this pull request still waiting on these changes? If so can I pick them up?

mesta1:
Sure, go on.

uknickwilliams:
How do I push my changes to submit a pull request? I am not allowed to push a new branch to the repository. Get a 403 error.

mesta1:
Fork the repo in your account, create a branch, push your code and create a commit, then you can create a PR.

mesta1:
Merged as part of https://github.com/killnine/s7netplus/pull/89

PR #89, opened 2017-08-31
Support atomic bit writing protocol, instead of read-modify-write

uknickwilliams:


mesta1:
Well done, thank you.

Issue #95, opened 2017-11-30
Reading UDInt or DWord with uint property doesn't read the correct value


thomasbargetzkuenz:
The plc DB looks like this:

![image](https://user-images.githubusercontent.com/30795774/33437981-e0669ac0-d5e9-11e7-89af-befe20a68673.png)

The class to read the DB into looks like this:

```csharp
    public class TestPoco
    {
        public bool BoolValue { get; set; }
        public byte ByteValue { get; set; }
        public short IntValue { get; set; }
        public ushort WordValue { get; set; }
        public int DIntValue { get; set; }
        public uint DWordValue { get; set; }
        public double RealValue { get; set; }
    }
```

Read result:

![image](https://user-images.githubusercontent.com/30795774/33438328-c6f59afe-d5ea-11e7-8a61-104e9f57b57d.png)

As you can see the read DWordValue is not correct. The value on the plc is in hex `AAAAD555` which is `2863322453` in decimal. The read decimal value is `1440066218`.

I think the problem is in `Class.GetPropertyValue(Type, bytes[], ref double)`:

```csharp
                case "Int32":
                    numBytes = Math.Ceiling(numBytes);
                    if ((numBytes / 2 - Math.Floor(numBytes / 2.0)) > 0)
                        numBytes++;
                    // hier auswerten
                    uint sourceUInt = DWord.FromBytes(
                        bytes[(int)numBytes + 3],
                        bytes[(int)numBytes + 2],
                        bytes[(int)numBytes + 1],
                        bytes[(int)numBytes + 0]);
                    value = sourceUInt.ConvertToInt();
                    numBytes += 4;
                    break;
                case "UInt32":
                    numBytes = Math.Ceiling(numBytes);
                    if ((numBytes / 2 - Math.Floor(numBytes / 2.0)) > 0)
                        numBytes++;
                    // hier auswerten
                    value = DWord.FromBytes(
                        bytes[(int)numBytes],
                        bytes[(int)numBytes + 1],
                        bytes[(int)numBytes + 2],
                        bytes[(int)numBytes + 3]);
                    numBytes += 4;
                    break;
```

When reading an Int32 the byte ordering is different than reading an UInt32. But if you check the `DWord.FromByteArray` method it looks like this:

```csharp
        public static UInt32 FromByteArray(byte[] bytes)
        {
            return FromBytes(bytes[3], bytes[2], bytes[1], bytes[0]);
        }
```

[...]

Issue #88, opened 2017-08-22
Read Class with Word and String


mikestromi:
Hello,

I have created a class with some words, real, and dwords and a string.
I try to read a class see picture for definitions
All numeric values are correct but the string remains null on read

Can anybody help?
![s7readstring](https://user-images.githubusercontent.com/31240932/29560606-cab0659a-8732-11e7-9e17-57cc53d7782b.png)

mesta1:
I think you have to read the string as an array of bytes and then convert it. In the last version it's possible to read arrays from ReadClass.

Issue #86, opened 2017-08-08
Can not read negative Values


Happe79:
Hello,

So far, thank you very much for that library. It helps me a lot.

But I have a problem reading negative values from DB.

E.g. I know that in the address area DB331.12.0 (DB331.Offset.Bit) is a values stored like -12 from Type "DINT", I get "-2147483648" using:

  S7.Net.Plc spsPLC = new S7.Net.Plc(CpuType.S7300, "100.0.0.0", 0, 2);

 object Result = spsPLC.Read(S7.Net.DataType.DataBlock, 331, iOffset, S7.Net.VarType.DInt, 1);
 long iNewValueInt = System.Convert.ToInt64(Result);
  sValue = "" + iNewValueInt;

return sValue;

If the values in the DB are positve it is working fine.

Do you may have an idea?

Thank you.

Stefan

mesta1:
System.Convert.ToInt64 is not correct.
You have to use the conversions provided by the library.

MoriLuca:
Sorry but i really cant get over this, i kindly ask for your help.

As far as i know, those are the Siemens Base data Types and the conversion that im tring to implement:

Siemens Int      ( 16bit )  -> DotNet Int16
Siemens UInt    ( 16bit )  -> DotNet UInt16
Siemens DInt    ( 32bit )  -> DotNet Int32
Siemens DUInt  ( 32bit )  -> DotNet UInt32
Siemens real     ( 32bit )  -> DotNet double     ( As i read from other post )
Siemens LReal   ( 64bit )  -> DotNet double    ( As i read from other post )

Im working on a proget based on Dot Net Core, building an asp net MVC Project.

While reading and casting a Siemens Int like this is working for positive and negative numbers:

```
object o = new object();
o = PLC.Read(S7.Net.DataType.DataBlock, 1, 0, S7.Net.VarType.Int, 1);
Int16 j = Convert.ToInt16(o);
```

For the double Int Negative Values this is not working :

```
object o;
o = .PLC.Read(S7.Net.DataType.DataBlock, 1, 4, S7.Net.VarType.DInt, 1);
Int32 j = Convert.ToInt32(o);
```

I read in some ohter post that i have to use the library methods for conversions, but i can not understand how to use them.
Sorry im a novice, but for the S7.Net.Conversion.ConvertToInt(), used for retrive negative numbers from DB, im expected to pass an uint value and i dont understand how to use it.

mesta1:
[...]

PR #101, opened 2018-03-18
fix data type converting for num-types

tomsoftware:
- fix data type converting
- add new string (stringEx) data type... my s7 returns on read

### DIFF
--- a/S7.Net/Types/DWord.cs
+++ b/S7.Net/Types/DWord.cs
@@ -12,16 +12,18 @@ namespace S7.Net.Types
         /// </summary>
         public static UInt32 FromByteArray(byte[] bytes)
         {
-            return FromBytes(bytes[3], bytes[2], bytes[1], bytes[0]);
+            return (UInt32)(bytes[0] << 24 | bytes[1] << 16 | bytes[2] << 8 | bytes[3]);
         }
 
+
         /// <summary>
-        /// Converts a S7 DWord (4 bytes) to uint (UInt32)
+        /// Converts 4 bytes to DWord (UInt32)
         /// </summary>
-        public static UInt32 FromBytes(byte v1, byte v2, byte v3, byte v4)
+        public static UInt32 FromBytes(byte b1, byte b2, byte b3, byte b4)
         {
-            return (UInt32)(v1 + v2 * Math.Pow(2, 8) + v3 * Math.Pow(2, 16) + v4 * Math.Pow(2, 24));
+            return (UInt32)((b4 << 24) | (b3 << 16) | (b2 << 8) | b1);
         }
+
 
         /// <summary>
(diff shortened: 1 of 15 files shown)


tomsoftware:
the original code uses: 
- byte to string to number converting
- byte shift by using pow(2, x)
this push uses
- byte to number converting by using bit operations

mesta1:
Thanks, I'll have a look. Is there any open issue that are fixed in these commits ?

tomsoftware:
- It fixes some float-point-problems in "Double.cs" for cases where the numbers a quiet small/big
- there were some cases with signed numbers where the result was wrong
- the new StringEx handels a String format my PLC returned on a read of string... don't know why

so may this fixes the  #95, #86 and #88 issue...

mesta1:
I really need unit tests on this. It seems all okay and tests are green, but I'm not sure all things that you changed are covered by tests. I will have a closer look on the weekend, but if you can write some tests, it will speed the merge.

S7.Net/Types/Timer.cs
```diff
@@ -13,26 +13,30 @@ public static class Timer
...
+            ///int unknown = (bytes[0] >> 6) & 0x03;
+
+            switch ((bytes[0] >> 4) & 0x03)
             {
-                case "00":
+                case 0:
                     wert *= 0.01;
                     break;
-                case "01":
+                case 1:
```
mbalous: Using binary literals might be more obvious 

```csharp
case 0b00:
    break;
case 0b01:
    break;
case 0b10:
    break;
case 0b11:
    break;
```

tomsoftware: Bit values are good for displaying FLAGs where every bit has a different meaning but in this case it is more or less a ENUM or even a exponentiation: ( wert *= 10^(n-2) ) so Int values are good I think?!

S7.Net/Types/DWord.cs
```diff
@@ -12,36 +12,39 @@ public static class DWord
...
+            return (UInt32)(bytes[0] << 24 | bytes[1] << 16 | bytes[2] << 8 | bytes[3]);
         }
 
+
         /// <summary>
-        /// Converts a S7 DWord (4 bytes) to uint (UInt32)
+        /// Converts 4 bytes to DWord (UInt32)
         /// </summary>
-        public static UInt32 FromBytes(byte v1, byte v2, byte v3, byte v4)
+        public static UInt32 FromBytes(byte b1, byte b2, byte b3, byte b4)
```
rapha-dev: Maybe more obvious naming is better when talking about bytes in a DWord/Word, starting at byte 0.
`public static UInt32 FromBytes(byte b0, byte b1, byte b2, byte b3)`

PR #102, opened 2018-03-21
Non breaking changes. Code styling, comments, etc...

mbalous:
Improved XML comments.
Throwing more descriptive exceptions.
Checking constructor parameters.
Improved code styling consistency.

These should be non-breaking changes.

### DIFF
--- a/S7.Net/PLC.cs
+++ b/S7.Net/PLC.cs
@@ -17,31 +17,32 @@ namespace S7.Net
     public class Plc : IDisposable
     {
         private const int CONNECTION_TIMED_OUT_ERROR_CODE = 10060;
-
-        private Socket _mSocket; //TCP connection to device
+        
+        //TCP connection to device
+        private Socket _mSocket;
 
         /// <summary>
-        /// Ip address of the plc
+        /// IP address of the PLC
         /// </summary>
         public string IP { get; private set; }
 
         /// <summary>
-        /// Cpu type of the plc
+        /// CPU type of the PLC
         /// </summary>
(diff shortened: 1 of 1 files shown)


S7.Net/PLC.cs
```diff
@@ -108,8 +108,14 @@ public bool IsConnected
...
         {
-            IP = ip;
+            if (!Enum.IsDefined(typeof(CpuType), cpu))
+                throw new InvalidEnumArgumentException(nameof(cpu), (int) cpu, typeof(CpuType));
+
+            if (string.IsNullOrEmpty(ip))
+                throw new ArgumentException("IP address must valid.", nameof(ip));
+
             CPU = cpu;
+            IP = ip;
```
mbalous: Field IP is public _get_ only. It would make sense to make it IPAddress directly.
```csharp
public string IP { get; private set; }
```
to
```csharp
public IPAddress IP { get; private set; }
```
And in constructor to

```csharp
IP = IPAddress.Parse(ip);
```

That would caught invalid IP right at the beginning.

mesta1:
Thank you ! I have to look at the previous PR, it will take a while, then I'll get this merged.

S7.Net/PLC.cs
mesta1: I prefere not to break anything, if possible. Imho wrong IP address is not a major concern. If it's invalid, it will throw exception.

mbalous: Ok, makes sense.

PR #107, opened 2018-04-06
Fix communication with WinAC/SoftPLC CPUs and Fix write double

thoj:
Hi

Thanks for this library! I stumbled on some issues when trying to write some values to a SIPROTECT unit (WinAC/LC). 

Issue1: All softPLC based units apparently send out an early "ACK" before most replies that messes with the code. This patch fixes it. The packet in question is "03-00-00-07-02-F0-00". I just wrote a function that trows away this packet if it sees it. 

Issue2: It seems the code for writing single values is broken for some types. When writing double it casts it to a UInt32 and throws away numbers after the decimal point and writes it as a Uint32 on the PLC. This patch fixes this for for double.

### DIFF
--- a/S7.Net/PLC.cs
+++ b/S7.Net/PLC.cs
@@ -169,11 +169,27 @@ namespace S7.Net
                 return LastErrorCode;
             }
 
-            try 
+            try
             {
-                byte[] bSend1 = { 3, 0, 0, 22, 17, 224, 0, 0, 0, 46, 0, 193, 2, 1, 0, 194, 2, 3, 0, 192, 1, 9 };
+                byte[] bSend1 = { 3, 0, 0, 22, //TPKT
+                    17,     //COTP Header Length
+                    224,    //Connect Request 
+                    0, 0,   //Destination Reference
+                    0, 46,  //Source Reference
+                    0,      //Flags
+                    193,    //Parameter Code (src-tasp)
+                    2,      //Parameter Length
+                    1, 0,   //Source TASP
+                    194,    //Parameter Code (dst-tasp)
+                    2,      //Parameter Length
+                    3, 0,   //Destination TASP
+                    192,    //Parameter Code (tpdu-size)
+                    1,      //Parameter Length
+                    9       //TPDU Size (2^9 = 512)
+                };
 
-                switch (CPU) {
+                switch (CPU)
+                {
                     case CpuType.S7200:
                         //S7200: Chr(193) & Chr(2) & Chr(16) & Chr(0) 'Eigener Tsap
                         bSend1[11] = 193;
--- a/S7.Net/PLC.cs
+++ b/S7.Net/PLC.cs
@@ -39,6 +39,11 @@ namespace S7.Net
         /// Slot of the CPU of the PLC
         /// </summary>
         public Int16 Slot { get; private set; }
+
+        /// <summary>
+        /// Max PDU size this cpu supports
+        /// </summary>
+        public Int16 MaxPDUSize { get; private set; }
         
         /// <summary>
         /// Returns true if a connection to the PLC can be established
@@ -158,7 +163,6 @@ namespace S7.Net
         /// <returns>Returns ErrorCode.NoError if the connection was successful, otherwise check the ErrorCode</returns>
         public ErrorCode Open()
         {
-            byte[] bReceive = new byte[256];
 
             _mSocket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
             _mSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReceiveTimeout, 1000);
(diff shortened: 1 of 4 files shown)


mesta1:
There were 2 PR that I had to merge. Can you please review your conflicts and make a new commit ?

thoj:
Rebased and squashed some stuff. Hope you can use it :)

S7.Net/PLC.cs
```diff
@@ -229,28 +366,33 @@ public ErrorCode Open()
                         return ErrorCode.WrongCPU_Type;
                 }
 
+                //COTP Setup
                 _mSocket.Send(bSend1, 22, SocketFlags.None);
-                if (_mSocket.Receive(bReceive, 22, SocketFlags.None) != 22)
+                var response = ReadCOTP_TPDU();
```
mesta1: Shouldn't check for 22 before calling ReadCOTP_TPDU ?

S7.Net/PLC.cs
```diff
@@ -229,28 +366,33 @@ public ErrorCode Open()
...
                     throw new Exception(ErrorCode.WrongNumberReceivedBytes.ToString());
-                } 
+                }
 
+                //S7ComSetup
                 byte[] bsend2 = { 3, 0, 0, 25, 2, 240, 128, 50, 1, 0, 0, 255, 255, 0, 8, 0, 0, 240, 0, 0, 3, 0, 3, 1, 0 };
                 _mSocket.Send(bsend2, 25, SocketFlags.None);
 
-                if (_mSocket.Receive(bReceive, 27, SocketFlags.None) != 27)
+                var responseLength = ReciveCOTP_TSDU(bReceive);
```
mesta1: Same here check for 27 before? or is it fine to go into the ReciveCOTP_TSDU

S7.Net/PLC.cs
```diff
@@ -277,10 +419,10 @@ public void ReadMultipleVars(List<DataItem> dataItems)
         {
             int cntBytes = dataItems.Sum(dataItem => VarTypeToByteLength(dataItem.VarType, dataItem.Count));
             
-            if (dataItems.Count > 20)
-                throw new Exception("Too many vars requested");
-            if (cntBytes > 222)
-                throw new Exception("Too many bytes requested"); // TODO: proper TDU check + split in multiple requests
+            //Is this really a limit?
```
mesta1: Did you try with more than 20 dataItems?

S7.Net/PLC.cs
```diff
@@ -277,10 +419,10 @@ public void ReadMultipleVars(List<DataItem> dataItems)
...
+            //if (dataItems.Count > 20) throw new Exception("Too many vars requested");
+            if (cntBytes > MaxPDUSize - 18)
+                throw new Exception("Too many bytes requested"); //TODO: split in multiple requests
```
mesta1: Add MaxPDUSize - 18 in the message, to show what is the maximum.

S7.Net/PLC.cs
```diff
@@ -297,11 +439,11 @@ public void ReadMultipleVars(List<DataItem> dataItems)
                 _mSocket.Send(package.array, package.array.Length, SocketFlags.None);
 
                 byte[] bReceive = new byte[512];
-                int numReceived = _mSocket.Receive(bReceive, 512, SocketFlags.None);
-                if (bReceive[21] != 0xff)
-                    throw new Exception(ErrorCode.WrongNumberReceivedBytes.ToString());
 
-                int offset = 25;
+                int numReceived = ReciveCOTP_TSDU(bReceive);
```
mesta1: Rec - e - iveCOTP_TSDU

S7.Net/PLC.cs
```diff
@@ -1113,8 +1257,8 @@ private ErrorCode WriteBytesWithASingleRequest(DataType dataType, int db, int st
 
                 _mSocket.Send(package.array, package.array.Length, SocketFlags.None);
 
-                int numReceived = _mSocket.Receive(bReceive, 512, SocketFlags.None);
-                if (bReceive[21] != 0xff)
+                int numReceived = ReciveCOTP_TSDU(bReceive);
+                if (bReceive[14] != 0xff)
```
mesta1: why did this change from 21 to 14 ?

S7.Net/PLC.cs
```diff
@@ -153,6 +159,121 @@ private ErrorCode Connect(Socket socket)
             return LastErrorCode;
         }
 
+        /// <summary>
+        /// Describes a TPKT Packet
+        /// </summary>
+        private class TPKT
```
mesta1: I'm not a fan of  private classes. Can you move these to internal classes in their own files ?

thoj: Agreed 😉

S7.Net/PLC.cs
thoj: Nope. I'm pretty sure I removed this check but it came back somehow. I'll fix 😊

S7.Net/PLC.cs
thoj: Yup 😣

S7.Net/PLC.cs
thoj: I did not try but I really can't find this limit anywhere else. Including in libnodave.

S7.Net/PLC.cs
thoj: Hehe woops 😫

S7.Net/PLC.cs
thoj: The reason is that cotp_tsdu.data only contains the S7comm data. Both headers for tpkt and cotp are removed.

thoj:
I did the changes. I only did a small test against a SoftPLC cpu. It seems to work. I'm out traveling so when I get home I can test against more CPUs.

The next step I think is to add parsing of the S7comm header and return an object instead of the byte[] array.

mesta1:
Great thank you

thoj:
Ah ok I'll test more when I get home.

Issue #115, opened 2018-04-26
Possible bug in ReadMultipleVars on VarType.Bit?


rapha-dev:
Is there a bug in ReadMultipleVars when a VarType.Bit is used?

```
var dataItems = new List<DataItem>()
{
	new DataItem
	{
		Count = 1,
		DataType = DataType.DataBlock,
		DB = 2,
		StartByteAdr = 0,
		BitAdr = 0,
		VarType = VarType.Bit
	},
	new DataItem
	{
		Count = 1,
		DataType = DataType.DataBlock,
		DB = 2,
		StartByteAdr = 0,
		BitAdr = 4,
		VarType = VarType.Bit
	},
	new DataItem
	{
		Count = 1,
		DataType = DataType.DataBlock,
		DB = 2,
		StartByteAdr = 16380,
		BitAdr = 7,
		VarType = VarType.Bit
	}
};
plc.ReadMultipleVars(dataItems);
```

Package capture Job:
```
TPKT, Version: 3, Length: 55
ISO 8073/X.224 COTP Connection-Oriented Transport Protocol
S7 Communication
    Header: (Job)
    Parameter: (Read Var)
        Function: Read Var (0x04)
        Item count: 3
        Item [1]: (DB2.DBX 0.0 BYTE 1)
            Variable specification: 0x12
            Length of following address specification: 10
            Syntax Id: S7ANY (0x10)
            Transport size: BYTE (2)
            Length: 1
            DB number: 2
            Area: Data blocks (DB) (0x84)
            Address: 0x000000
        Item [2]: (DB2.DBX 0.0 BYTE 1)
            Variable specification: 0x12
            Length of following address specification: 10
            Syntax Id: S7ANY (0x10)
            Transport size: BYTE (2)
            Length: 1
            DB number: 2
            Area: Data blocks (DB) (0x84)
            Address: 0x000000
        Item [3]: (DB2.DBX 16380.0 BYTE 1)
            Variable specification: 0x12
            Length of following address specification: 10
            Syntax Id: S7ANY (0x10)
            Transport size: BYTE (2)
            Length: 1
            DB number: 2
            Area: Data blocks (DB) (0x84)
            Address: 0x01ffe0

```

Pcap Ack_Data:
```
TPKT, Version: 3, Length: 38
ISO 8073/X.224 COTP Connection-Oriented Transport Protocol
S7 Communication
    Header: (Ack_Data)
    Parameter: (Read Var)
        Function: Read Var (0x04)
[...]

PR #117, opened 2018-04-27
Fixed bug in ReadMultipleVars on VarType.Bit

rapha-dev:
Also added BitAdr to Types.DataItem to use Bits in ReadMultipleVars.
Expanded existing UnitTest.
Refer to #115.

### DIFF
--- a/S7.Net.UnitTest/S7NetTests.cs
+++ b/S7.Net.UnitTest/S7NetTests.cs
@@ -386,19 +386,51 @@ namespace S7.Net.UnitTest
         {
             Assert.IsTrue(plc.IsConnected, "Before executing this test, the plc must be connected. Check constructor.");
 
-            ushort val = 16384;
-            plc.Write("DB2.DBW16384", val);
-            ushort result = (ushort)plc.Read("DB2.DBW16384");
-            Assert.AreEqual(val, result, "A ushort goes from 0 to 64512");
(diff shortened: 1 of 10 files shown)


S7.Net.UnitTest/S7NetTests.cs
```diff
@@ -407,19 +430,40 @@ public void T10_ReadMultipleBytes()
...
                     VarType = VarType.Word
-                }
+                },
+                new DataItem
+                {
+                    Count = 1,
+                    DataType = DataType.DataBlock,
+                    DB = 2,
+                    StartByteAdr = 16384,
+                    BitAdr = 6,
```
mesta1: This is undefined.

mesta1:
Can you do the merge?

rapha-dev:
#108 changed PLC response data a little, had to re-test.
Merged to master correctly.

S7.Net.UnitTest/S7NetTests.cs
rapha-dev: Added BitAdr to S7.Net/Types/DataItem.cs, was lost in previous merge.

mycroes:
Tested this, seems to work fine. Without this change all results following a single byte read are offset by one byte.

mesta1:
Cool thank you

PR #118, opened 2018-04-28
Fix miscalculation of ClassSize & StructSize

GrayDelacluyse:
Currently if certain datatypes are used, for example boolean you can end up where the remainder is 0.125.  When casting to int, it will truncate and miscalculate the number of bytes to read.


mesta1:
This was fixed in a previous PR.

PR #121, opened 2018-05-25
Add multiple write support

mycroes:
This is my initial draft to support write of multiple variables using a single request. The implementation is based on the implementation in Sharp7, but actually supporting value conversion and using infrastructure from S7NetPlus.

It's currently lacking some checks (like validating data size), but it does work. It actually supports write of separate bits as well. I guess it might only work for the DataBlock area, as I didn't focus on any other areas. Also, I didn't bother to add xmldoc yet either.

I intend to make additional changes to this, so I'm not considering this done, but I'd love to get some feedback early on.

### DIFF
--- a/S7.Net/PLCHelpers.cs
+++ b/S7.Net/PLCHelpers.cs
@@ -316,42 +316,6 @@ namespace S7.Net
             }
         }
 
-        public byte[] GetPackage(object value)
-        {
-            switch (value.GetType().Name)
-            {
-                case "Byte":
-                    return Types.Byte.ToByteArray((byte)value);
-                case "Int16":
-                    return Types.Int.ToByteArray((Int16)value);
-                case "UInt16":
-                    return Types.Word.ToByteArray((UInt16)value);
-                case "Int32":
-                    return Types.DInt.ToByteArray((Int32)value);
-                case "UInt32":
-                    return Types.DWord.ToByteArray((UInt32)value);
-                case "Double":
-                    return Types.Double.ToByteArray((double)value);
-                case "Byte[]":
-                    return (byte[])value;
-                case "Int16[]":
-                    return Types.Int.ToByteArray((Int16[])value);
-                case "UInt16[]":
-                    return Types.Word.ToByteArray((UInt16[])value);
-                case "Int32[]":
-                    return Types.DInt.ToByteArray((Int32[])value);
-                case "UInt32[]":
-                    return Types.DWord.ToByteArray((UInt32[])value);
-                case "Double[]":
-                    return Types.Double.ToByteArray((double[])value);
-                case "String":
-                    return Types.String.ToByteArray(value as string);
-                default:
-                    throw new InvalidVariableTypeException();
-            }
-        }
-
-
         /// <summary>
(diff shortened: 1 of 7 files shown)


S7.Net/PLCHelpers.cs
```diff
@@ -172,6 +172,27 @@ private void ParseString(string address)
 
     public partial class Plc
     {
+        public static ushort GetWordAt(IList<byte> buf, int index)
```
mesta1: Please move static helpers outside plc class.

S7.Net/PLCHelpers.cs
```diff
@@ -316,10 +337,12 @@ private object ParseBytes(VarType varType, byte[] bytes, int varCount, byte bitA
             }
         }
 
-        public byte[] GetPackage(object value)
+        public static byte[] GetPackage(object value)
```
mesta1: Please move static helpers outside plc class.

S7.Net/Protocol/S7WriteMultiple.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using S7.Net.Types;
+
+namespace S7.Net.Protocol
+{
+    public static class S7WriteMultiple
+    {
+        public static int CreateRequest(byte[] pdu, DataItem[] dataItems)
```
mesta1: This seems to be used only internally. If that's the case, mark the class and method as internal.

mesta1:
Also it would be nice if you create unit tests for the new features

S7.Net/PLCHelpers.cs
mycroes: Will do. Also, this breaks the API (even just making it static does), let me know how you feel about this. Also, it would probably make sense to have this internal as well.

S7.Net/Protocol/S7WriteMultiple.cs
mycroes: Yes it is. I'm refactoring this to improve readability/performance, will keep in mind to make it internal at some point.

S7.Net/PLCHelpers.cs
mycroes: Will do.

S7.Net/PLCHelpers.cs
mesta1: I think this is used only internally. What we shouldn't break are the methods used externally, the ones in the documentation.
Also we should really mark this helpers methods for the protocol as internal.

mycroes: I agree, but nevertheless it breaks API. I'm not sure if you're already using semantic versioning, but that would seem like a good idea to me, meaning that changes like this will increment the major version number.

mycroes:
Updated the code to conform more to the rest of s7netplus. Probably could be refactored to use shared code to construct TPKT, ISO DT and S7 Job Request, but I think it's OK in current form. I intend to still add unit tests, however I have verified that this does work, both against an Siemens S7-1500 PLC and against Snap7 running as server.

PR #123, opened 2018-06-05
Port to NetStandard

deinok:
For now the support is only for netstandard2.0

More will be done in a future PR to port also to netstandard1.X


mesta1:
Hi, thanks for the PR but it's still too early for .Net standard 2.0. 
For now I'll keep it with old .Net, but I'm open to change it once old versions of .Net Framework (4.5 - 4.6) will phase out.

mycroes:
This is actually changing the project to multitarget (`<TargetFrameworks>net452;netstandard2.0</TargetFrameworks>`), which means it targets both .NET 4.5.2 (and Visual Studio will treat it as such because `net452` is listed first) and NetStandard 2. I'd say this is a welcome change, also because the project file is easier to work with.

mesta1:
The problem is that the IDE is not supporting multi targeting, so unless you know about multi targeting and manually edit the csproj, you have the project locked. 
Since it's really easy to move to .Net standard 2, I'm not willing to do the change now. Again in the future I will consider that, especially if Microsoft fixes the IDE.

deinok:
@mesta1I dont actualy see the problem. Can you explain it better?

mycroes:
@deinok I think [this](https://weblog.west-wind.com/posts/2017/Jun/22/MultiTargeting-and-Porting-a-NET-Library-to-NET-Core-20) is what @mesta1 is referring to...

deinok:
image

deinok:
So, lazynes is the problem?
As I can see, this project dont need precompilation directives or anything else in order to work well in multiple frameworks.

mycroes:
@deinok I don't think it has anything to do with lazyness. On a sidenote, I'm working on [my own S7 connection library](https://github.com/mycroes/Sally) as well, which in fact is a NetStandard project. It's far from done, but it does implement some basic reading and writing with speed similar to S7NetPlus, so perhaps it can fulfill your needs.

deinok:
Well, im also implementing my own. Its focused on speed, not easy to work with. Actualy I have TPKT and parts of COTP done.

mycroes:
@deinok I just released the first version of mine: [nuget.org / Sally7](https://www.nuget.org/packages/Sally7/0.1.0), feel free to give it a test run. Also, since we're both focusing on speed it would be interesting to compare our libraries.

deinok:
Interesting, but as I'd said, just TPKP is ready for production in my case. :(

Seems like you are playing with Spans like me : D

I will give it a try tomorrow ;)

mycroes:
As of #136, I'm considering this for merging. @rapha-dev @thoj please weigh in on this.

rapha-dev:
I like the idea to have it all in one project, personally I don't care whether it is configurable via UI or manual file changes, it should be documented though.
So, a GO from me!

deinok:
Yeah, with this change we can have the same simplificated csproj for netframework, netstandard, core and UWP.
The S7.Net.Core package could be deleted after some fix on the code after this change

thoj:
This looks good to me. As long as it multitarget this change is fine by me.

deinok:
@thoj Yes, now targets net452 like before, and also netstandard2.0

thoj:
I tested this patch and it works fine for me.

mycroes:
Thank you @deinok for the patch and thank you @rapha-dev and @thoj for feedback. I hope you guys will keep this up.

@deinok I don't have any UWP experience, so without investigating myself can you tell me what needs to be done to supersede the .Core project? Want to create an issue for it to track requirements and progress?

Issue #124, opened 2018-06-07
ErrorCode.ReadData never returned


arkypita:
Hi
I see that calling a plc.Read(...) will set a lasterror of "ErrorCode.WriteData". Should not return ErrorCode.ReadData?

I have search for ErrorCode.ReadData reference in code, seem to be only declared but not used.

Suggest:  ReadBytesWithSingleRequest in PlcSyncronous.cs should set ErrorCode.ReadData, not WriteData

rapha-dev:
Agreed, same should be done in `public async Task<List<DataItem>> ReadMultipleVarsAsync(List<DataItem> dataItems)` and `public void ReadMultipleVars(List<DataItem> dataItems)`.

BTW: Why is it marked as obsolete? In my understanding Sync and Async should be exist both unless it is used internally.

mycroes:
> BTW: Why is it marked as obsolete? In my understanding Sync and Async should be exist both unless it is used internally.

Things that can be done `async` should be done `async`. If a user choses not to use `async`, he should make that decision at the top-most level where he can make the decision. As such it does make sense to obsolete the sync API (even though Microsoft will never do such a thing because of backwards compatibility). Note that it's not my decision though, it was changed in PR #111.

mesta1:
About the Obsolete, I didn't notice that. Can someone remove that please? I don't think we will ever remove the synchronous version. @mycroes @rapha-dev

mycroes:
@mesta1 Will do.

mycroes:
It was actually only there on ReadMultipleVars and in some comments, removed all mentions of obsolete in PR #129.

rapha-dev:
Fixed in #125  and #129, may gets obsolete in #137.

PR #125, opened 2018-06-08
Fixes for Exception naming ReadData

rapha-dev:
As suggested in #124.

### DIFF
--- a/S7.Net/PLCHelpers.cs
+++ b/S7.Net/PLCHelpers.cs
@@ -154,7 +154,7 @@ namespace S7.Net
                             varType = VarType.Counter;
                             return;
                         default:
-                            throw new InvalidAddressException(string.Format("{0} is not av valid address", address.Substring(0, 1)));
+                            throw new InvalidAddressException(string.Format("{0} is not a valid address", address.Substring(0, 1)));
                     }
 
                     string txt2 = address.Substring(1);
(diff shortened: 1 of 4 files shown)


S7.Net/PLCHelpers.cs
```diff
@@ -248,7 +248,7 @@ private ByteArray CreateReadDataRequestPackage(DataType dataType, int db, int st
         /// <returns></returns>
         private object ParseBytes(VarType varType, byte[] bytes, int varCount, byte bitAdr = 0)
         {
-            if (bytes == null)
+            if (bytes == null || bytes.Length == 0)
```
rapha-dev: Check for empty byte[] here before it gets past Converter functions (and raise unhandled exceptions), returned from ReadBytes.
```
                byte[] bytes = ReadBytesWithSingleRequest(dataType, db, index, maxToRead);
                if (bytes == null)
                    return resultBytes.ToArray();
```

S7.Net/PLCHelpers.cs
```diff
@@ -352,10 +352,10 @@ public byte[] GetPackage(object value)
...
 
 
-    /// <summary>
-    /// Sets the <see cref="LastErrorCode"/> to <see cref="ErrorCode.NoError"/> and <see cref="LastErrorString"/> to <see cref="string.Empty"/>.
-    /// </summary>
-    public void ClearLastError()
+        /// <summary>
+        /// Sets the <see cref="LastErrorCode"/> to <see cref="ErrorCode.NoError"/> and <see cref="LastErrorString"/> to <see cref="string.Empty"/>.
+        /// </summary>
+        public void ClearLastError()
```
rapha-dev: Beautification only.

PR #133, opened 2018-06-22
Calculate PDU size constraints

mycroes:
This calculates PDU size for multiple read/write. I didn't bother to include single read / write calculation yet (but it would be fair if you'd expect me to add that before merging). I tested this against the S7-1500 PLC I have access to, and it **didn't** work. By that I mean that I wasn't able to read 20 dataItems (19 is max supported on 240 byte PDU, but negotation results in 960 byte PDU), so I'm not sure what's broken yet. I haven't tried this with Sally7 yet, so I'm not sure where the issue is. Nevertheless I wanted to start the PR for this.

The calculations are actually based on information from the SAPI-S7 manual, I'm assuming correctness of the manual to this regard (it does seem to make sense, although I'm not sure where they start counting towards the PDU size, at least TPKT and COTP DT seem to be excluded).


rapha-dev:
Have tested on a S7-300 (via Simulator), max PDU should be 240 but negotiated is 480 which won't work, and gives an exception as expected when reading 20 DataItems.
It looks like there is something broken in negotiation (or Simulator is fooling me). Will have a test on a real S7-300 as well, have no S7-400 available ATM but another S7-1500.

rapha-dev:
It looks like #145 doesn't affect reading 20 (different) DataItems, however reading 2000 bytesw as array works.

rapha-dev:
Yes, I know the Siemens SAPI manual. But there is no evidence nor description that limits a read request to 240 bytes without beeing affected by negotiated PDU size. The returned data bytes can be likely PDU size.

However, splitting requests might be take some more effort, as you noted.

mycroes:
Right, there's no limit at 240 bytes. With Sally7 I can read more than 20 variables, so it's definitely an issue with S7NetPlus. I haven't investigated any further though, but most certainly the constraint of 19 variables should only be in case of 240 byte PDU size.

rapha-dev:
I see and maybe found the issue, it is not about contraints yet, but fixes the Socket exception while reading > 19 DataItems.

bauerju:
Hi @mycroes 
Have an additional approach for read and write, It would split the List into small portions which fit into pdu size (according to your AssertPduSizeForRead / AssertPduSizeForWrite

S7Data is an enhanced DataItem. It holds its ByteArraySize in value NumBytes.  

        public void SplitWrite(List<S7Data> s7Datas)
        {
                var toWrite = new List<S7.Net.Types.DataItem>();

                int cnt = 0;
                int bytes = 0;

                foreach (var s7Data in s7Datas.OrderBy(x => x.dataItem.DB).ThenBy(x => x.dataItem.StartByteAdr).ToList())
                {
                    s7Data.OldValue = s7Data.Value;

                    if (12 + cnt * 18 + 18 > S7Plc.MaxPDUSize  // 12 bytes of header data, 18 bytes of parameter data for each dataItem, + 12 for next item! 
                      | 12 + cnt * 16 + bytes + s7Data.NumBytes + 16 > S7Plc.MaxPDUSize) // 12 bytes of header data, 16 bytes of data for each dataItem and the actual data + 16 for next item! 
                    {
                        Write(toWrite.ToArray());
                        bytes = 0; cnt = 0;
                        toWrite.Clear();
                    }
                    toWrite.Add(s7Data.dataItem);
                    cnt += 1;
                    bytes += s7Data.NumBytes;
                }

                if (toWrite.Count > 0)
                    Write(toWrite.ToArray());

                toWrite.Clear();
            }
        }
 
and the read:

      public void SplitRead(List<S7Data> s7Datas)
        {

                var toRead = new List<S7.Net.Types.DataItem>();

                int cnt = 0;
                int bytes = 0;

                foreach (var s7Data in s7Datas.OrderBy(x => x.dataItem.DB).ThenBy(x => x.dataItem.StartByteAdr).ToList())
                {
                    s7Data.OldValue = s7Data.Value;

                    if (12 + cnt * 12 + 12 > S7Plc.MaxPDUSize   // 12 bytes of header data, 12 bytes of parameter data for each dataItem + 12 for next item!
                      | 14 + cnt * 4 + bytes + s7Data.NumBytes + 4 > S7Plc.MaxPDUSize) // 14 bytes of header data, 4 bytes of result data for each dataItem and the actual data + 4 for next item!
                    {
                        ReadMultipleVars(toRead);
                        bytes = 0; cnt = 0;
                        toRead.Clear();
                    }
                    toRead.Add(s7Data.dataItem);
                    cnt += 1;
                    bytes += s7Data.NumBytes;
                }

                if (toRead.Count > 0)
                    ReadMultipleVars(toRead);
                toRead.Clear();

              
        }

bauerju:
BTW I tested against a S7/1507.
MaxPdu was here 960 and reading worked with the max possible 79 Items .

Didn't check against an S7 300

Jürgen

mycroes:
Hi @bauerju,

Thanks for your comments and tests. I'll keep your ideas for splitting reads / writes in mind. I have always taken the position that the developer needs to know when splitting happens, but I can also imagine that in many situations it doesn't matter (also, for my work it often doesn't matter since we take special care about the order of written data and always using a command/status field to indicate data was written/received).

We're planning on making some structural changes to S7NetPlus, that could also include this behavior, perhaps configurable by the user.

bauerju:
Hi @mycroes 
Yes developer should know whe splitting happens. But sometimes it is not necesarry to know. I also have some bits for handshake with the plc to signalice that the written data is valid. So I write two times: first the data ( here the order doesn't care), later the handshake. I think ever develeoper should do this in an equivalent way.

Let me ask, which structural changes do you plan? I'm curious. 

Jürgen

mycroes:
HI @bauerju,

As far as structure is concerned, points we discussed are:
 - Creating an interface for Plc (which will ease the addition of derivative implementations)
 - Possibly making DataItem generic
 - Reducing the API surface (some methods can be used to perform identical actions, but the API generally isn't very clear)
 - Fixing terminology (which might not have much impact after the other changes)

So functionality shouldn't really be impacted, but you might have to change some calls / arguments. Also, these are plans, AFAIK noone has started on them yet (@killnine is actually getting involved again, he had some ideas with regards to the structure).

killnine:
Are you talking about splitting up large reads/writes into chunks that fit within the PDU limitations? Or talking about splitting reads/writes so that if you write you don't immediately do a subsequent read to verify?

Perhaps we should take this discussion into Slack or create a different issue card for it as this is getting pretty long.

bauerju:
hi @killnine 
the above coding splits large reads/writes into chunks. It tries to fit pdu size.  

Jürgen

mycroes:
Closing to perform against develop.

PR #141, opened 2018-06-30
Corrected NuGetPack.bat for create desired targets only

rapha-dev:
This patch creates a new version directory like `nuget-pack/lib/0.1.9/` and put all build contents into, including `xml` documentation.

Should we keep `nuget-pack/lib/` directory in git?

### DIFF
--- a/NuGetPack.bat
+++ b/NuGetPack.bat
@@ -1,12 +1,16 @@
 @echo off
-set /p version="Version: "
+REM Ask for Release version unless supplied by first commandline arg
+if "%1" == "" (
+	set /p version="Version: "
+) else (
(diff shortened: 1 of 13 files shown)


mycroes:
We should avoid having generated / compiled files in the repository. Derek (@killnine) and I think that the NuGet package should be generated from AppVeyor, either automatically or with a simple human action. I guess we should also create an organization on NuGet to tidy this all up.

I'd rather see a PR removing all binaries from the repo than this PR, as that is what's considered good Git practice.

rapha-dev:
That sounds better than manually generating NuGet packages, so just removing binaries from git should be sufficient.

PR #142, opened 2018-06-30
Support for NetStandard 1.3

mycroes:
This adds support for NetStandard 1.3 and drops the old UWP solution and S7.Net.Core project. I haven't actually tested this on UWP, it's just that it compiles. Also, due to the lack of features in NetStandard 1.3 I'm not sure if the struct and class reading will behave exactly identical. Major difference is that there's no `Type.GetProperties()` in NetStandard 1.3, so I used `Type.DeclaredProperties`. Honestly I didn't bother to check, but I guess it might only return properties declared on the given type and not inherited properties (that it doesn't actually declare). I remember seeing a solution for that somewhere, where `DeclaredProperties` from `Type.BaseType` are also included (recursively), at least that's what I think.

If someone can test this that would be nice. @deinok can you review this as well?


deinok:
Hi, this seems a real rework.

Taking a look at : https://docs.microsoft.com/es-es/dotnet/standard/net-standard#net-implementation-support

Actually we support the netstandard2.0 and also net452, so the runtime we dont support is netcoreapp1.X

So probably supporting first netstandard1.6 we could avoid most of the lacks of APIs you found while supporting netcoreapp1.X

So I would like most of this PR but in more small PRs. For example
- Cleanup (S7.Net.Core)
- Add netstandard16 (IDK if it needs more things than just adding it to targetframeworks)
- TcpClient Compat
- COTP rework
- Add netstandard1.3 and try to avoid using precompilation directives

mycroes:
> Actually we support the netstandard2.0 and also net452, so the runtime we dont support is netcoreapp1.X

We don't support UWP 10.0, which is the only target in S7.Net.Core. I actually started by adding `uap10.0` to TargetFrameworks, but it's not well supported right now and it had similar missing API's.

> So probably supporting first netstandard1.6 we could avoid most of the lacks of APIs you found while supporting netcoreapp1.X

That's a false truth. Yes it could avoid some missing API's, but no, my target was not .NET core 1.0. Also, if you know you need to target a lower NetStandard version, it doesn't make much sense to target a higher version first.

> So I would like most of this PR but in more small PRs. For example

Except for the cleanups this is all tightly related work for a single feature, splitting this in smaller PR's doesn't make sense to me.

> Add netstandard1.3 and try to avoid using precompilation directives

I don't see much of an issue with precompiler directives. We want support for net452 so the highest NetStandard version we could use as common denominator would be 1.2. I'm not sure about the state of TcpClient there, but either way it doesn't have System.Runtime.Serialization. We could opt to use NetStandard 1.3 as replacement for NetStandard 2.0, but that would still leave the issues mentioned in the PR description.

I was hoping you could give some technical feedback, since I never wrote an UWP app.

deinok:
So, why do you need something lower or exact than 1.3.
Before 1.3 there is  no support  for anything related to sockets i think

mycroes:
@deinok NetStandard 1.3 or 1.4 is needed for UWP 10.0. Only UWP 10.0.16299 and up support NetStandard 1.5, 1.6 and 2.0. I'm honestly not sure if there's any target audience there (i.e., with UWP 10.0 lower than 10.0.16299), but I just added this to simplify the project as a whole while retaining the same level of compatibility (also, I'm not sure if the .NET Core build actually works right now).

As far as .NET Core is concerned, with NetStandard 1.3 we actually support .NET Core 1.0, but I guess I'm even less concerned about people being stuck on .NET Core 1.0 versus people being stuck on UWP 10.0.

NetStandard 1.3 also allows use with all Xamarin platforms, but it is supported by the same platform versions that also support NetStandard 1.6 (so no negative impact there if we'd target NetStandard 1.6).

Now about NetStandard 1.6, it would only bring us a `TcpClient` that is compatible with .NET Framework / NetStandard 2.0. All the other stuff (some reflection, and `System.Runtime.Serialization`) is missing there just as well, so it wouldn't change much.

deinok:
Okey, just checked it with netstandard1.6. Seems like the bigger problem is SerializationInfo , GetBuffer, and a TcpClose.
Just the same as netstandard1.3

I'd just want to avoid the use of `NET_FULL` but that will break backward compat.
So, LGTM

rapha-dev:
Please also update Readme.md regarding target frameworks once this is approved. Can I test something against (real) S7-300?

mycroes:
I started working on a UWP sample. It's actually working and it's insanely ugly. I would normally use ReactiveUI to build UI apps, however ReactiveUI doesn't support UWP 10.0 (which I chose as target so that I can actually assert that the NetStandard 1.3 build works). Anyway, it connects, reads, writes, but that's it.

mycroes:
Removed UWP sample from this PR. The sample has proven it's working, merging the PR.

Issue #132, opened 2018-06-22
s7netplus incorrectly uses `double` as the .NET type for REAL values in PLC


mycroes:
`Double` values in .NET are 8 bytes long, `float` / `Single` are 4 bytes as is the `REAL` type in Siemens PLC's. Also, except for byte ordering the bitwise values are equal.

@mesta1 My proposal would be to mark all `double` related conversions as obsolete and replace them with `float` conversions. I'm not sure what would be sane way to handle double fields in structs/classes, my personal opinion would be to throw an exception on those and up the s7netplus version to 2.0.0.

rapha-dev:
I'd like to vote +1 on this as I had some kind of misunderstanding before.

rapha-dev:
Have prepared the implementation of float / Single, needs some testing against real hardware but should work. Will also create a PR for this.

mycroes:
Fixed by #143.

PR #143, opened 2018-07-02
Add support for datatype float (Single) and obsolete double (Double) …

rapha-dev:
…usage.
Expanded all unit tests.
Addresses #132.


S7.Net.UnitTest/S7NetTestsAsync.cs
```diff
@@ -815,6 +839,19 @@ public partial class S7NetTests
             Assert.AreEqual(test_value, test_value2, 0.01, "Compare Write/Read"); //Need delta here because S7 only has 32 bit reals
         }
 
+        [TestMethod]
+        public async Task Test_Async_ReadWriteSingle()
+        {
+            float test_value = 55.6632f;
+            await plc.WriteAsync("DB1.DBD0", test_value);
+            Assert.AreEqual(plc.LastErrorCode, ErrorCode.NoError, "Write Single");
```
mycroes: The third parameter for `Assert` should be a message (i.e. "After WriteAsync Plc.LastErrorCode should be 'NoError', but LastErrorCode is '{plc.ErrorCode}'."). I guess this isn't applied anywhere though ;-)

S7.Net/Conversion.cs
```diff
@@ -204,23 +204,49 @@ public static UInt32 ConvertToUInt(this int input)
         /// </summary>
         /// <param name="input"></param>
         /// <returns></returns>
-        public static UInt32 ConvertToUInt(this double input) 
+        [Obsolete("This method will soon be deprecated. Use ConvertToUInt(float) instead.")]
```
mycroes: `[Obsolete("Double support is obsolete. Use ConvertToUInt(float) instead.")]`

S7.Net/Conversion.cs
```diff
@@ -204,23 +204,49 @@ public static UInt32 ConvertToUInt(this int input)
...
+            output = S7.Net.Types.DWord.FromByteArray(S7.Net.Types.Single.ToByteArray(input));
+            return output;
+        }
+
         /// <summary>
         /// Converts from DWord (DBD) to double
         /// </summary>
         /// <param name="input"></param>
         /// <returns></returns>
+        [Obsolete("This method will soon be deprecated. Use ConvertToFloat(uint) instead.")]
```
mycroes: "Double support is obsolete. ...

S7.Net/Protocol/Serialization.cs
```diff
@@ -8,39 +8,43 @@ internal static class Serialization
...
-                    return Types.DInt.ToByteArray((Int32[]) value);
+                    return Types.DInt.ToByteArray((Int32[])value);
                 case "UInt32[]":
-                    return Types.DWord.ToByteArray((UInt32[]) value);
+                    return Types.DWord.ToByteArray((UInt32[])value);
                 case "Double[]":
-                    return Types.Double.ToByteArray((double[]) value);
+                    return Types.Double.ToByteArray((double[])value);
+                case "Single[]":
+                    return Types.Single.ToByteArray((float[])value);
```
mycroes: I'm wondering what you're using for the formatting, because I guess the default R# settings will actually have a space between type and value in a cast.

S7.Net.UnitTest/S7NetTestsAsync.cs
rapha-dev: Yeah, UnitTests needs a review, I may start on this.

S7.Net/Protocol/Serialization.cs
rapha-dev: Nope, my VS doesn't do that. Spacing after cast is disabled and never touched it (so is [default setting](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference#spacing)). Look at TSDU class, there aren't spaces after cast either, so i guess my setting is commonly used.

rapha-dev:
@mycroes 
Thanks for your input and suggestions, do you have a S7-400 available for testing? Not only asking for this PR, in general.

mycroes:
No, I only have access to S7 1200 and S7 1500.

PR #144, opened 2018-07-02
UWP sample project

mycroes:
UWP sample project, currently work in progress.


mycroes:
It seemed like a good idea 4 years ago, but I don't think UWP has much value left in it. Abandoning...

Issue #145, opened 2018-07-03
ReadBytesWithSingleRequest cannot read >491 Bytes on S7-1500


rapha-dev:
Possible bug in COTP CR when connecting to a S7-1500, TPDU is too small (actually 512 bytes). Any Read resulting in `actualPDU + COTPHeader > TPDU` (which is 957 + 3 = 960 > 512) will fail (Socket disconnects).

Maximum Bytes of 491 would work: 509 + 3 = 512 > 512

For reference: [PLC4X](https://plc4x.incubator.apache.org/protocols/s7/s7comm.html)

Increasing the TPDU to maximum of 1024 bytes (still need to check on S7-300, 400 and 1200) will allow to read an unspecified maximum of bytes as PDUs will be split into multiple PDUs with maximum size.

I will create a PR for this and need some testing.

mycroes:
This probably also relates to open issue for PR #133.

> Increasing the TPDU to maximum of 1024 bytes

We're negotiating 1920 byte PDU, so maybe we should increase to 2048. Also, the _ISO over TCP_ spec states the default is **65355** bytes IIRC, which means we could possibly zero out the PDU size entirely. We should really have access to all supported PLC's to verify these changes though.

rapha-dev:
You are right, I have made some tests using different sizes up to 2^15, using 2^11 should be sufficient to support max PDU on all available PLCs.

rapha-dev:
Merged into `develop`.

PR #146, opened 2018-07-03
ReadBytesWithSingleRequest cannot read >491 Bytes on S7-1500

rapha-dev:
Addressed issue #145.

Tested against PLC types:
- [ ] S7-200
- [x] S7-300
- [ ] S7-400
- [ ] S7-1200
- [x] S7-1500
- [x] Snap7
- [x] PlcSim

@mycroes Can you please test against live S7 1200?

Forgot about S7-200, anyone using it?


mycroes:
I'm not sure when I will have time to test with the S7-1200, so I might merge this without testing.

PR #152, opened 2018-07-04
Fixes invalid TPKT length in request header

rapha-dev:
TPKT length field is actually 2 bytes, so use them correctly.

### DIFF
--- a/S7.Net/PLCHelpers.cs
+++ b/S7.Net/PLCHelpers.cs
@@ -177,9 +177,9 @@ namespace S7.Net
         {
             //header size = 19 bytes
             var package = new Types.ByteArray(19);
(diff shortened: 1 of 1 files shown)


mycroes:
Good catch, this probably wouldn't have been the first place I would've checked.

mycroes:
@rapha-dev Did you get my emails btw?

rapha-dev:
@mycroes 
No, did not get mail from you, where have you sent it?
As you said issue is not in Sally7 I fired up a quick example and compared chats using Wireshark, so thanks for that.

mycroes:
@rapha-dev I sent it to the address that's used on your commits, from the address that's used in my commits.

rapha-dev:
I see, that mail is not reachable, may have to change it. Have sent a mail to you.

PR #156, opened 2018-07-07
Replace LastErrorCode and LastErrorString with exceptions

### LINKED ISSUE #137: Exception vs LastErrorCode / LastErrorString
As the title says, use of `Exception`s vs `LastErrorCode` and `LastErrorString`. Personally I favor exceptions because they don't break under concurrent use and they're the builtin (as in within the language) solution for this problem. A custom exception type can be created to also return the error code. There are some cases where exceptions are thrown on precondition failures while in other cases the error code and message are set, I think this should be implemented consistently.

mycroes:
Fixes #137.

### DIFF
--- a/S7.Net/PlcSynchronous.cs
+++ b/S7.Net/PlcSynchronous.cs
@@ -11,16 +11,12 @@ namespace S7.Net
     public partial class Plc
     {
         /// <summary>
-        /// Open a <see cref="Socket"/> and connects to the PLC, sending all the corrected package
-        /// and returning if the connection was successful (<see cref="ErrorCode.NoError"/>) of it was wrong.
+        /// Connects to the PLC and performs a COTP ConnectionRequest and S7 CommunicationSetup.
         /// </summary>
-        /// <returns>Returns ErrorCode.NoError if the connection was successful, otherwise check the ErrorCode</returns>
-        public ErrorCode Open()
+        public void Open()
         {
-            if (Connect() != ErrorCode.NoError)
-            {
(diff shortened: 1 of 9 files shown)


S7.Net/PlcSynchronous.cs
```diff
@@ -547,15 +503,9 @@ public void ReadMultipleVars(List<DataItem> dataItems)
```
rapha-dev: Some Exceptions are still generic, should be `PlcException` here.

mycroes:
If you hover over a line there's be a blue box with a plus near the start of the line, click it to comment inline.

mycroes:
Fixed the 3 issues you pointed out.

To be honest this is not exactly how I would normally do exception handling, but this is closest to the behavior with the `LastErrorCode` and `LastErrorString`.

rapha-dev:
It might be better to commit this PR to a new `development` branch since it will break existing projects without having a proper exception handling. It was necessary to monitor `ErrorCode` to ensure everything went well, now it is gone.

mycroes:
I agree, keep in mind though that the only difference the develop branch will make is that it's easier to release hotfixes for the current codebase. I'm not sure if I can change the default branch, but my proposal would be to change the default branch to develop and merge only releases to master (taking GitFlow as branching model).

rapha-dev:
Agreed, GitFlow also sounds good.

rapha-dev:
Closing this as it would merge to master, replaced by #161

PR #162, opened 2018-07-09
Fix write stringex

### LINKED ISSUE #159: Incorrect String handling with vartyp StringEX
Hi. Writing a string over a daitem, declared as StringEX,  doesn't work. 

The first two bytes in then memory, which represents declared and used length are overwriten.  
Simple Code:
2018-07-09 07_54_40-flsng - microsoft visual studio

DB before writing:
2018-07-09 07_43_25-clipboard

DB after writing:
2018-07-09 07_45_11-s7 watch

Jürgen

mycroes:
Fixes #159.

### DIFF
--- a/S7.Net.UnitTest/S7.Net.UnitTest.csproj
+++ b/S7.Net.UnitTest/S7.Net.UnitTest.csproj
@@ -78,6 +78,7 @@
     <Compile Include="S7NetTestsSync.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Helpers\TestLongStruct.cs" />
+    <Compile Include="TypeTests\StringExTests.cs" />
   </ItemGroup>
   <ItemGroup>
     <Content Include="snap7.dll">
--- a/S7.Net.UnitTest/TypeTests/StringExTests.cs
+++ b/S7.Net.UnitTest/TypeTests/StringExTests.cs
@@ -1,0 +1,115 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using S7.Net.Types;
+
+namespace S7.Net.UnitTest.TypeTests
+{
+    [TestClass]
+    public class StringExTests
+    {
+        [TestMethod]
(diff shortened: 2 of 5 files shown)


mycroes:
@bauerju could you elaborate on that? I think you take that as providing the string length to this method, not the length including header.

bauerju:
Hi @mycroes 
Here my litte Test Programm: 
           var da = new S7.Net.Types.DataItem()
            {
                DataType = S7.Net.DataType.DataBlock,
                Count = 16,
                DB = 1100,
                BitAdr = 0,
                VarType = S7.Net.VarType.StringEx,
                Value = "Test"
            };
            var plc = new S7.Net.Plc(cpu: S7.Net.CpuType.S7300, ip: "172.30.215.235", rack: 0, slot: 2);
            plc.Open();
            plc.Write(da);


for Iine 36 
`ToByteArray` is called here with lengthIncludingHeader = 16 . Should this not be 18? 

In function PlcHelpers.VarTypeToByteLength you add 2 to the count parameter for StringEx. So I thought count declares the string length not the used byte length in plc. 

for line 47 
You declare the byte with  length +2. So you can only write length bytes from position 2 into the array.

Jürgen

mycroes:
I honestly didn't check what was used elsewhere, but I'll change it to match the length calculation. Will probably be fixed this evening. Thanks for the review!

mycroes:
@bauerju Fixed the conversion, should work as intended now. About line 47 (or 46 now):
```csharp
bytes[1] = (byte) Encoding.ASCII.GetBytes(value, 0, reservedLength, bytes, 2);
```
`reservedLength` is the maximum amount of characters, so we don't need to use `value` here. The method will return the amount of characters copied, so that should be fine as well. the `2` near the end is the offset in the byte array where the characters will be written to. I will add a unit test to ensure functionality.

mycroes:
Not sure if I should say I verified the code, since it was actually broken... Tests now indicate that it's working as expected, so either both the tests and implementation are wrong or it's actually working.

PR #163, opened 2018-07-10
Add SourceLink.Copy.PdbFiles to add missing .pdb files

mycroes:
The PDB files are the missing part to allow debugging in VS without having local sources for S7NetPlus.


killnine:
I think we should register release PDBs, if they're really necessary, to a symbol server and not check them into source.

mycroes:
@killnine They're not included in source control, they're included in the NuGet package. They're using a modern technique where there's a reference to the source files on GitHub in the pdf, Visual Studio will download those files on demand. See [this section](https://github.com/ctaggart/SourceLink#package-pdb-in-nupkg) for some more information.

killnine:
Ah, slick. I misunderstood what you were doing.

PR #167, opened 2018-07-11
Fix incorrect length when reading BitArray

### LINKED ISSUE #165: Reading bits - length parameter not the same
If I read bits with this statement:
`var bits = (BitArray) plc.Read(DataType.DataBlock, 1, 1, VarType.Bit, 2, 0);`

Than I get 16 bits in a BitArray.

If I read bits with this:
`var bits = (BitArray) plc.Read(DataType.DataBlock, 1, 1, VarType.Bit, 1, 0);`

I just get a single bit (true/false) With or without casting to Bitarray makes no difference.

mycroes:
Fixes #165

### DIFF
--- a/S7.Net/Types/Bit.cs
+++ b/S7.Net/Types/Bit.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections;
+﻿using System;
+using System.Collections;
 
 namespace S7.Net.Types
 {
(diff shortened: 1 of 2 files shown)


S7.Net/Types/Bit.cs
```diff
@@ -16,12 +17,27 @@ public static bool FromByte(byte v, byte bitAdr)
...
+        /// <returns>A BitArray with the same number of bits and equal values as <paramref name="bytes"/>.</returns>
+        public static BitArray ToBitArray(byte[] bytes) => ToBitArray(bytes, bytes.Length * 8);
+
+        /// <summary>
+        /// Converts an array of bytes to a BitArray.
+        /// </summary>
+        /// <param name="bytes">The bytes to convert.</param>
+        /// <param name="length">The number of bits to return.</param>
+        /// <returns>A BitArray with <paramref name="length"/> bits.</returns>
+        public static BitArray ToBitArray(byte[] bytes, int length)
```
rapha-dev: A `offset` is needed here, when a different starting bit other than 0 is used it returns beginning at bit 0.
`var result = (BitArray) plc.Read(DataType.DataBlock, 800, 0, VarType.Bit, 2, 7);`
Result shows `DB800.DBX0.0` to `DB800.DBX0.1` not `DB800.DBX0.7` to `DB800.DBX1.0` as expected, inner request shows 2 bytes, that's correct.

mycroes: We should verify that having an offset is possible, because I think arrays **always** start at bit 0. Not verified though. Also, in that case we should throw on `bitAdr != 0`.

mycroes: I have to disagree. I wouldn't want to give the false intention to users that byte arrays from a different bit than 0 are supported. We could trick reading them (but as far as we know, the protocol doesn't support reading with a bit offset), but we can never write them. I guess a common way that people will use S7N+ is to read a `DataItem` and to write it back, logically the user shouldn't need to create 2 `DataItems` to do that.

Let me know what you think, I'm strongly in favor of only accepting bitAdr 0 when reading arrays.

rapha-dev: True when it comes to writing. You are also right when using a `DataItem` it can be a single bit to read/write it. However we can add a converter method to extract a `BitArray` from byte(s), or users have to implement itself.

Issue #150, opened 2018-07-03
Inconsistent specification of bit index


mycroes:
Bit index (or `bitAdr` in it's most common form in S7NetPlus) has inconsistent values applied as default throughout the code. In some case it defaults to `-1`, in others it defaults to `0` (from what I remember).

Proposal, replace `bitAdr` with `bitIndex` and change it's type to `int?` (nullable) and default the value to `null`.

PR #169, opened 2018-07-12
Dataitem from string write fix

### LINKED ISSUE #168: DataItem created from DataItem.FromAddress() fail to write with error 5 in branch: development
Hi
I'm testing latest development branch:  
I created a bunch of DataItems with `DataItem.FromAddress(string)`. 

Writing with `Write(DataItem[])` fails with Exception:  
> Write failed for 3 items. See the innerExceptions for details.

InnerException is

> Write of dataItem S7.Net.Types.DataItem failed with error code 5.

Problem seems to be the BitAdr of the DataItems.
It is converted from the result of bitnumber -1 in `PLCAddress.Parse` to 255 in DataItem.BitAdr.
If I manually set this to 0 for all VarTypes excluding Bit. then the write gives no error.

Jürgen

mycroes:
Fixes #168, references #150.

### DIFF
--- a/S7.Net/Protocol/S7WriteMultiple.cs
+++ b/S7.Net/Protocol/S7WriteMultiple.cs
@@ -30,13 +30,13 @@ namespace S7.Net.Protocol
                 Serialization.SetWordAt(message, paramOffset + Parameter.Offsets.Amount, (ushort) value.Length);
                 Serialization.SetWordAt(message, paramOffset + Parameter.Offsets.DbNumber, (ushort) item.DB);
                 message[paramOffset + Parameter.Offsets.Area] = (byte) item.DataType;
-                Serialization.SetAddressAt(message, paramOffset + Parameter.Offsets.Address, item.StartByteAdr, item.BitAdr);
-
-                paramOffset += Parameter.Template.Length;
 
                 data.Add(0x00);
                 if (item.Value is bool b)
                 {
+                    Serialization.SetAddressAt(message, paramOffset + Parameter.Offsets.Address, item.StartByteAdr,
+                        item.BitAdr);
+
                     data.Add(0x03);
                     data.AddWord(1);
 
@@ -45,6 +45,8 @@ namespace S7.Net.Protocol
                 }
                 else
                 {
+                    Serialization.SetAddressAt(message, paramOffset + Parameter.Offsets.Address, item.StartByteAdr, 0);
+
                     var len = value.Length;
                     data.Add(0x04);
                     data.AddWord((ushort) (len << 3));
@@ -55,8 +57,9 @@ namespace S7.Net.Protocol
                         data.Add(0);
                     }
                 }
-            }
 
+                paramOffset += Parameter.Template.Length;
+            }
 
             message.Add(data.Array);
 
--- a/S7.Net/Protocol/S7WriteMultiple.cs
+++ b/S7.Net/Protocol/S7WriteMultiple.cs
@@ -34,6 +34,10 @@ namespace S7.Net.Protocol
                 data.Add(0x00);
                 if (item.Value is bool b)
                 {
+                    if (item.BitAdr > 7)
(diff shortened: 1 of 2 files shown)


bauerju:
Hi @mycroes 
This seems to work. Thank you.

But I searched for a non obvious behaviour:
Writing  a byte to DBB1 worked only the first time. Second write went to position  DBB4.
Here my source : 

            var dc = S7.Net.Types.DataItem.FromAddress("DB1100.DBB1");
            var plc = new S7.Net.Plc(cpu: S7.Net.CpuType.S7300, ip: "172.30.215.235", rack: 0, slot: 2);
            plc.Open();
            dc.Value = (byte)1;

            plc.Write(new S7.Net.Types.DataItem[] { dc });

            plc.ReadMultipleVars(new List<S7.Net.Types.DataItem>() { dc });

            dc.Value = (byte)dc.Value + 1; // -> value gets converted to int !

            plc.Write(new S7.Net.Types.DataItem[] { dc });

            plc.ReadMultipleVars(new List<S7.Net.Types.DataItem>() { dc });

Problem was: Byte was converted to integer in source by adding the 1. And as DataItem does not really care about setted Vartype the integer was later mapped instead of a byte to a array of 4 bytes . 

This makes using it dangerous. In my application I use a wrapper around DataItem which only allows setting of correct type or converts it if possible.  

May bee all other write methods have the same behavior. Didn't test it yet. 

Jürgen

rapha-dev:
@bauerju 
We have also discovered that issue, fix is on the way.

bauerju:
Hi @rapha-dev 
I uses this 


      public object Value
        {
            get => (dataItem.VarType != VarType.StringEx && dataItem.VarType != VarType.String) ? dataItem?.Value :
                 ((string)dataItem.Value)?.Split('\0')[0]; // Bei String nur den Teil bis \0 ausgeben

            set => dataItem.Value = Convert.ChangeType(value, UsedDataType);
        }

UsedDataType is set from VarType :

      Type UsedDataType;
      private void CalcByteCount(int count, out int numBytes, out Type usedDataType)
        {
            usedDataType = null;
            numBytes = 1;

         
       switch (dataItem.VarType)
            {
                case VarType.Bit:
                    numBytes = count;
                    usedDataType = count == 1 ? typeof(bool) : typeof(bool[]);
                    break;

                case VarType.Byte:
                    if (count >= 1)
                    {
                        numBytes = count;
                    }
                    numBytes = 1;
                    usedDataType = count == 1 ? typeof(byte) : typeof(byte[]);
                    break;

                case VarType.Word:
                    usedDataType = count == 1 ? typeof(UInt16) : typeof(UInt16[]);
                    numBytes = count * 2;
                    break;

                case VarType.Int:
                    usedDataType = count == 1 ? typeof(Int16) : typeof(Int16[]);
                    numBytes = count * 2;
                    break;

                case VarType.Timer:
                    usedDataType = count == 1 ? typeof(UInt16) : typeof(UInt16[]);
                    numBytes = count * 2;
                    break;

                case VarType.Counter:
                    usedDataType = count == 1 ? typeof(UInt16) : typeof(UInt16[]);
                    numBytes = count * 2;
                    break;

                case VarType.DWord:
                    usedDataType = count == 1 ? typeof(UInt32) : typeof(UInt32[]);
                    numBytes = count * 4;
                    break;

                case VarType.DInt:
                    usedDataType = count == 1 ? typeof(Int32) : typeof(Int32[]);
                    numBytes = count * 4;
                    break;

                case VarType.Real:
                    usedDataType = count == 1 ? typeof(Single) : typeof(Single[]);
                    numBytes = count * 4;
                    break;

                case VarType.String:
                    usedDataType = typeof(string);
                    numBytes = count;
                    break;

                case VarType.StringEx:
                    usedDataType = typeof(string);
                    numBytes = count + 2;
                    break;
            }
        }

Jürgen

killnine:
Thanks for testing.

PR #174, opened 2018-08-09
Implement optional timeout parameter in the OpenAsync function.

mbalous:
This PR consists of two commits. 

1. Added optional parameter to function `OpenAsync`.
2. Created test for this function when the parameter is specified.

New method signatures:
```csharp

public partial class Plc
{
    public async Task OpenAsync(int timeout = 0);
    private async Task ConnectAsync(int timeout);
}
```

### DIFF
--- a/S7.Net.UnitTest/S7NetTestsAsync.cs
+++ b/S7.Net.UnitTest/S7NetTestsAsync.cs
@@ -48,6 +48,32 @@ namespace S7.Net.UnitTest
             }
         }
 
+        [TestMethod]
+        public async Task Test_Async_Connection_Timeout()
+        {
+            // According to https://tools.ietf.org/html/rfc5737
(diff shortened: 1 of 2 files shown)


S7.Net.UnitTest/S7NetTestsAsync.cs
```diff
@@ -48,6 +48,32 @@ public async Task Test_Async_Connection()
...
+            sw.Start();
+            try
+            {
+                await unavailablePlc.OpenAsync(50);
+            }
+            catch (System.TimeoutException ex)
+            {
+                sw.Stop();                
+                Assert.IsInstanceOfType(ex, typeof(System.TimeoutException));
+                if (sw.ElapsedMilliseconds > 500)
```
mycroes: Not sure why it **has** to be thrown within 500ms?

PR #177, opened 2018-08-14
 Support multi thread call synchronous methods in one instance.

NiHangBoy:
Support multi thread call synchronous methods in one instance.

### DIFF
--- a/S7.Net/PlcSynchronous.cs
+++ b/S7.Net/PlcSynchronous.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net.Sockets;
 using S7.Net.Protocol;
+using S7.Net.Thread;
 
 //Implement synchronous methods here
 namespace S7.Net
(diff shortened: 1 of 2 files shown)


mycroes:
Hi @NiHangBoy,

First of all, thank you for your contribution. I have a couple of things to say:
- There are quite some issues with this code. I recommend you read some more about thread safety (for instance, your code isn't thread-safe, it might be somewhat thread-safe for 2 threads only from what I see).
- There is no reason `Plc` _needs_ to be thread-safe. I understand you would _like_ it to be thread-safe, but I don't want to force locking for everyone. In a future S7NetPlus release there will be an `IPlc` (or similarly named) interface, that would allow for a thread-safe counterpart to the current implementation without forcing it on everyone. Even then, locking might not be the way to go (you're blocking a potential unlimited amount of threads now with the `AutoResetEvent.WaitOne()`).
- You can easily wrap S7NetPlus calls using [`ExclusiveTaskQueue`](https://gist.github.com/mycroes/c4f23b056494fa844ad071f1c10782ca). That would probably scratch your itch, although it doesn't provide the uttermost performance because it synchronizes access on a higher level than ultimately required. On the other hand I can hardly imagine it being an actual problem, I'm using similar techniques for forcing synchronized access as well. Using this method also prevents thread blocking by the way.

Long story short, I don't see this change fit for inclusion and I don't see a current path where we want to include this. When structural changes will come to S7NetPlus, there will definitely be a thread-safe counterpart for `Plc`.

PR #189, opened 2018-09-21
Add DateTime support

### LINKED ISSUE #187: Request: Support DATE_AND_TIME complex type
Hello,
I should read DATE_AND_TIME complex type:
http://www.plccenter.cn/Siemens_Step7/Zusammengesetzte_Datentypen.htm
would be nice if the library supported reading these into C# DateTime, or does the library have some way of doing this already?

mycroes:
Fixes #187.

### DIFF
--- a/S7.Net.UnitTest/S7.Net.UnitTest.csproj
+++ b/S7.Net.UnitTest/S7.Net.UnitTest.csproj
@@ -79,6 +79,7 @@
     <Compile Include="S7NetTestsSync.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Helpers\TestLongStruct.cs" />
+    <Compile Include="TypeTests\DateTimeTests.cs" />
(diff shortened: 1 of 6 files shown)


FalcoGoodbody:
Hello @mycroes ,

I´m using S7NetPlus since the beginning of this year and it works really fine! Thank you for your work so far.

Today I need to read and write a DateTime value from and to the PLC and I am wondering how to do that? In my version of S7NetPlus I cannot find a datatype "DateTime" in the object-cataloge in VS2017 (I installed it via nuget) and I also cannot find any documentation to read or write a datetime-value.

Can you or someone else give me a hint how I can solve my little problem?

Thanks in advance!

mycroes:
Hi @FalcoGoodbody,

Sorry for my ultra late response. This PR wasn't merged, because I'm not sure if it actually works as it should. I'll merge it anyway and will prepare a new release, you can use it when the release is on NuGet.

Issue #190, opened 2018-09-25
Addresses are not aligned in case of custom type in class


admo:
Hi,
I've found that in file s7netplus/S7.Net/Types/Class.cs when class has property of type bool, then the numBytes are aligned in every case, but not in custom type.

This leads to read wrong addresses for properties of nested class. E.g.
```cs
class NestedProperties
{
  pubic Int16 Int16Prop{ get; set; };
}

class MyProperties
{
  public bool BoolProp { get; set; }
  public NestedProperties Prop2 { get; set; }
}

plc.ReadClass(new MyProperties(), 0, 0)
``` 
In above case `BoolProp` will be read from address `DB0.DBX0.0`, and `Int16Prop` from `DB0.DBW0`, but should be read from `DB0.DBW2`.

Please see:
https://github.com/S7NetPlus/s7netplus/blob/fba3ffd5dbe55ecf1ad739913a1fd955e906137a/S7.Net/Types/Class.cs#L64
and
https://github.com/S7NetPlus/s7netplus/blob/fba3ffd5dbe55ecf1ad739913a1fd955e906137a/S7.Net/Types/Class.cs#L205

I have proposet solution for this issue in PullRequest https://github.com/S7NetPlus/s7netplus/pull/191

PR #191, opened 2018-09-26
Fix for Addresses are not aligned in case of custom type in class

admo:
This is fix for issue https://github.com/S7NetPlus/s7netplus/issues/190

Treat inner custom class type as standard properties, when calculating address.

This PR contains implementation of the fix and new unit tests as well. All tests passed.


mycroes:
HI @admo, thanks for your contribution. I took a quick peek at the code, which seems to be OK (to be entirely honest I'm not impressed, but the class reading/writing is a bit of mess so I don't expect any changes to that to be pretty). Could you fix the spelling in the second commit (**writting** -> **writing**)? If that's fixed I'll take another good look but I expect to merge it then.

admo:
Hi @mycroes,
Spelling fixed.

mycroes:
@admo Thanks for your contribution. I'll try to get this into a release shortly.

PR #220, opened 2019-05-29
 Arrays must always start at even position 

jfontsaballs:
Tested with S7-1200.


mycroes:
Hi @jfontsaballs,
Not sure what you did, but there shouldn't be any merges in here ;-) Also, it's based on an older version of develop, make sure you fetch the latest from this repository first.

mycroes:
Addressed in #279.

PR #223, opened 2019-06-19
Fix padding of last dataitem in WriteMulitple fixes #222

### LINKED ISSUE #222: Write Byte / Bit DataItem failes with error code 33153 on 410-h5
Hi

This code fails on 410-5h cpu:

```
var di = new DataItem() { DataType = DataType.Memory, StartByteAdr = 0, VarType = VarType.Byte, Value = (byte)0xff, Count = 1 };
// Also fails:  var di = new DataItem() { DataType = DataType.Memory, StartByteAdr = 0, VarType = VarType.Bit, Value = true, Count = 1 };
            var foo = new DataItem[] { di };
            var plc = new Plc(CpuType.S7400, "192.168.3.211", 0, 5);
            plc.Open();
            plc.Write("MB0", (byte)0x00); //Works fine 
            plc.Write("MB1", (byte)0x00);
            plc.Write("MB2", (byte)0x00);
            plc.Write("MB3", (byte)0x00);           

            plc.Write(foo); //Write failed with error 33153.

            var v = plc.Read("MB0");
            Console.WriteLine(v);
            Console.Read();
            plc.Close();
```

This works: 

`var di = new DataItem() { DataType = DataType.Memory, StartByteAdr = 0, VarType = VarType.Byte, Value = 0xff, Count = 1 };`

This code works but writes to the wrong address. It writes to M3.0 for some reason. Seems like only writing Dint works properly? 
image

I'll try to debug more tomorrow. But any hints on what could be wrong?

thoj:
This fixes writing bits/bytes with WriteMultiple. It seems that CPU is very sensitive to extra padding with bytes/bits.

### DIFF
--- a/S7.Net/Protocol/S7WriteMultiple.cs
+++ b/S7.Net/Protocol/S7WriteMultiple.cs
@@ -20,8 +20,11 @@ namespace S7.Net.Protocol
             var dataOffset = paramOffset + paramSize;
             var data = new ByteArray();
 
+            var itemCount = 0;
+
             foreach (var item in dataItems)
             {
+                itemCount++;
                 message.Add(Parameter.Template);
                 var value = Serialization.SerializeDataItem(item);
                 var wordLen = item.Value is bool ? 1 : 2;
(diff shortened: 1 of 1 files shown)


mycroes:
Can you update the PR?

PR #225, opened 2019-07-10
Add Property to selected Port from PLC

tatzemax:
With this modification you can select the port of the PLC if it is behind a port forwarding.

### DIFF
--- a/S7.Net/PLC.cs
+++ b/S7.Net/PLC.cs
@@ -15,6 +15,9 @@ namespace S7.Net
         private TcpClient tcpClient;
         private NetworkStream stream;
 
+        private int readTimeout = System.Threading.Timeout.Infinite;
+        private int writeTimeout = System.Threading.Timeout.Infinite;
+
         /// <summary>
         /// IP address of the PLC
         /// </summary>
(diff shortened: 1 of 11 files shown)


S7.Net/PLC.cs
```diff
@@ -23,6 +23,11 @@ public partial class Plc : IDisposable
         /// </summary>
         public string IP { get; private set; }
 
+        /// <summary>
+        /// PORT Number of the PLC, default is 102
+        /// </summary>
+        public int PORT { get; private set; }
```
mycroes: Please rename to `Port`.

S7.Net/PLC.cs
```diff
@@ -107,7 +112,34 @@ public bool IsConnected
...
+        /// You need slot > 0 if you are connecting to external ethernet card (CP).
+        /// For S7-300 and S7-400 the default is rack = 0 and slot = 2.
+        /// </summary>
+        /// <param name="cpu">CpuType of the PLC (select from the enum)</param>
+        /// <param name="ip">Ip address of the PLC</param>
+        /// <param name="port">Port address of the PLC, default 102</param>
+        /// <param name="rack">rack of the PLC, usually it's 0, but check in the hardware configuration of Step7 or TIA portal</param>
+        /// <param name="slot">slot of the CPU of the PLC, usually it's 2 for S7300-S7400, 0 for S7-1200 and S7-1500.
+        ///  If you use an external ethernet card, this must be set accordingly.</param>
+        public Plc(CpuType cpu, string ip,int port, Int16 rack, Int16 slot)
```
mycroes: 1. I don't mind an optional port argument, but I don't think everyone should pass it by default.
2. Space missing before type.

tatzemax: 1. In line 154, the original function still exists.
2. is Fixed

S7.Net/PLC.cs
tatzemax: Fix

PR #244, opened 2019-12-18
Address parsing bug

timverwaal:
I made a small adjustment to the `PLCAddress.Parse` method. Because when parsing address M2000.1 the method returned `varType = Byte`, while I was expecting `Bit`.

I changed this in the code and added a test for this.
Also, attached a screenshot of the online Siemens configuration. Here you can see that M2000.1 is really a bit.
Screenshot


mycroes:
Hi Tim, it seems your branch is a bit off since it includes a few release commits as well. Can you change it to a single commit on top of current develop?

PR #245, opened 2019-12-18
Changed varType of memory Bit in address parsing method

timverwaal:
I made a small adjustment to the `PLCAddress.Parse` method. Because when parsing address M2000.1 the method returned `varType = Byte`, while I was expecting `Bit`.

I changed this in the code and added a test for this.
Also, attached a screenshot of the online Siemens configuration. Here you can see that M2000.1 is really a bit.
Screenshot

### DIFF
--- a/S7.Net.UnitTest/PLCAddressParsingTests.cs
+++ b/S7.Net.UnitTest/PLCAddressParsingTests.cs
@@ -1,0 +1,25 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using S7.Net.Types;
+using System;
+
+namespace S7.Net.UnitTest
+{
+    [TestClass]
+    public class PLCAddressParsingTests
+    {
+        [TestMethod]
+        public void T01_ParseM2000_1()
+        {
+            DataItem dataItem = DataItem.FromAddress("M2000.1");
+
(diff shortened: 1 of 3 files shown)


mycroes:
Hi Tim, I'm not a PLC programmer, but my gut feeling tells me that the varType for `M2000` (no dot, no index) should still be `Byte`. Are you able to verify this for me?
Regards,
Michael

timverwaal:
I'm not a PLC programmer either, but I have access to a test system for this.

When using `Datatype = Memory` and `varType = Byte`, the address will start with `MB`. This can be seen in the screenshot of my previous post for variables: `System_Byte` and `Clock_Byte`. I doublechecked this by creating a new variable `MtestVariable` with datatype `Byte` (refer to screenshot below, the error message in red is telling us we should use the B).

However, the code for this is already present in the `Parse` method. See code snippet below:

```C#
case "MB":
	// Memory byte
	dataType = DataType.Memory;
	dbNumber = 0;
	address = int.Parse(input.Substring(2));
	varType = VarType.Byte;
	return;
```

ScreenshotMTestVariable

timverwaal:
Upon further testing I noticed that more address types are missing.

image

I will add them to the `Parse` method and to the tests and then create a new pull request. 
For now I will close this one.

PR #258, opened 2020-03-08
Fixed received data size checking to avoid IndexOutOfRange exception

zoltraks:
This fixes a problem when trying to read DB from a PLC which has "Permit access with PUT/GET communication from remote partner" option in "Protection & Security" unchecked.

### DIFF
--- a/S7.Net/PlcSynchronous.cs
+++ b/S7.Net/PlcSynchronous.cs
@@ -353,7 +353,7 @@ namespace S7.Net
                 stream.Write(package.Array, 0, package.Array.Length);
 
                 var s7data = COTP.TSDU.Read(stream);
-                if (s7data == null || s7data[14] != 0xff)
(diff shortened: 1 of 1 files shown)


jakob-ledermann:
In the method [ReadBytesWithSingleRequestAsync in the file PlcAsynchronous.cs](https://github.com/S7NetPlus/s7netplus/blob/2fbf6591361dd2d879cbdccc70a9afa0ad6662c8/S7.Net/PlcAsynchronous.cs#L377) is the exact same check without your extension. It would be nice to have both cases fixed with the same PR.

mycroes:
@zoltraks Thanks for your contribution. I took note of @jakob-ledermann 's comment and addressed this issue in #282. Unfortunately this won't 'fix' the problem (that when PUT/GET is blocked you get an exception), but at least it will provide better error messages.

PR #267, opened 2020-06-11
Add support for DTL type in S71200 and S71500

Jason-Jelks:
DT Format is not natively supported by S71200 and S71500. It was changed to DTL (DateTimeLong). However S7200, S7300 & S7400 currently only support DT type format. I will add the required parsing to translate Windows DateTime.UTC to and from S71200/1500.

Additionally, the "Class" functions do not support DateTime, of any type (DT, DTL), to be part of the parsed Class structure. This can be supported by adding DI to the parsing methods for the instanced PLC.

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -4,14 +4,18 @@ branches:
   master:
     tag: rc
     increment: Minor
-  features?[/-]:
+  feature:
+    regex: features?[/-]
     tag: rc-{BranchName}
     increment: Minor
-  (pull|pull\-requests|pr)[/-]:
+  pull-request:
+    regex: (pull|pull\-requests|pr)[/-]
     tag: rc-pr-{BranchName}
     increment: Minor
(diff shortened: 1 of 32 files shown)


mycroes:
Hi @Jason-Jelks, sorry for this super late reply. I don't see any changes while I'd guess you must have made some. Do you want to give this another try?

mycroes:
Obsoleted by #284.

Jason-Jelks:
My apologies, I was traveling and unable to get back to this. It appears that #284 made very similar changes to what I did for the DateTimeLong. Thanks for picking up the ball!

mycroes:
No worries, thanks for your response!

Jason-Jelks:
FWIW - I did review #284 and merged some of his changes to mine, but left some of my changes for optimizations. I also have changes to the Class.cs which you can review and see if they work for you or not.

PR #279, opened 2020-08-13
Types/Class: Start arrays on even bytes

mycroes:
Addresses #175. Implements array offset fix only, char support is not included.


mycroes:
Also addresses #220, which basically implements the same behavior (without taking bits into consideration).

Issue #283, opened 2020-08-14
Unnecessary fail after Read (TKPT incomplete)


jakob-ledermann:
This might be related to #250 and simliar errors.
The current Implementation assumes, that a stream.Read only returns less than requested if the message is incomplete or a EOF is encountered. This assumption does not hold. NetworkStream.Read is guaranteed to block until the first byte is available and then copies the available bytes up to the requested count into the provided buffer.
If you are in a heavy load network environment (in my case 5 Mbit/s from a different device on the same network adapter + S71200) it happens that the Read function only copies a small amount of bytes into the buffer, if you call read again with an adjusted offset and the count reduced by the already received count, you often get more bytes of the expected message.

The current implementation (especially 0.4.0) fails after the first try. I hope i can get my fix up to 0.5.0 soon, so I can submit a PR.
One of the hard parts is to make sure I don't introduce an potentially endless loop.

mycroes:
Hi @jakob-ledermann, I figured this wouldn't happen because the message size is smaller than a normal MTU, so I didn't handle it in Sally7 either. It's definitely something that's worth fixing, so unless you absolutely want to tackle this yourself I'll take a jab at it.

mycroes:
Possibly related: #262 .

scamille:
https://stackoverflow.com/a/24412022 provides an example for a Stream ReadExactly extension function, which might be all that is needed here. Just needs to replace all the Stream.Read calls with it then. And a async version should be easy to add as well.

mycroes:
Thanks @scamille . I was expecting to fix it along the same lines, didn't have time to do something with it yet. I'll try to fix this this evening.

jakob-ledermann:
My internal solution, that only needs a sorting of the repositories:
`
        /// <summary>
        /// Reads a fixed amount of bytes from the stream into the buffer
        /// </summary>
        /// <param name="stream">the Stream to read from</param>
        /// <param name="buffer">the buffer to read into</param>
        /// <param name="offset">the offset in the buffer to read into</param>
        /// <param name="count">the amount of bytes to read into the buffer</param>
        /// <param name="timeoutMs">The timeout to abort the read</param>
        /// <returns>returns the amount of read bytes</returns>
        /// <exception cref="TimeoutException">
        ///  Throws timeout exception, when the timeout elapsed, while waiting for available Data
        /// </exception>
[...]

PR #288, opened 2020-08-18
Unittests for issue #283

jakob-ledermann:
These unittests fail intentionally as the issue discussion has not reached
a consesual solution. Any solution should pass the unittests added in this commit.

I intentionally request only the unittest as they should be mergable quite quickly.

### DIFF
--- a/S7.Net.UnitTest/ProtocolTests.cs
+++ b/S7.Net.UnitTest/ProtocolTests.cs
@@ -35,6 +35,33 @@ namespace S7.Net.UnitTest
         }
 
         [TestMethod]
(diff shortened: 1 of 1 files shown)


mycroes:
I can't merge these because AppVeyor rejects when tests fail. Let's keep this PR open until the solution is here as well.

jakob-ledermann:
Ok. As the tests need a little adjustment to be solvable anyway. Memorystreams seem to move EOF on Write around.

mycroes:
This was covered by #291, so I'm closing this now. Feel free to re-open if you think my assessment is invalid.

PR #289, opened 2020-08-18
Fixed size reads

jakob-ledermann:
solves #283.

Submitted as draft and open for editing so you can build upon my code. I hope we can achieve an agreement on the course of solution quickly.

I propose further discussions are moved from the issue to this place.

### DIFF
--- a/S7.Net.UnitTest/ProtocolTests.cs
+++ b/S7.Net.UnitTest/ProtocolTests.cs
@@ -10,6 +10,8 @@ using System.Threading.Tasks;
 
 namespace S7.Net.UnitTest
 {
+    using System.Threading;
+
     [TestClass]
     public class ProtocolUnitTest
     {
@@ -21,7 +23,7 @@ namespace S7.Net.UnitTest
             Assert.AreEqual(0x03, t.Version);
             Assert.AreEqual(0x29, t.Length);
             m.Position = 0;
-            t = TPKT.ReadAsync(m).Result;
+            t = TPKT.ReadAsync(m, CancellationToken.None).Result;
             Assert.AreEqual(0x03, t.Version);
             Assert.AreEqual(0x29, t.Length);
         }
@@ -66,7 +68,7 @@ namespace S7.Net.UnitTest
         public async Task TPKT_ReadShortAsync()
         {
             var m = new MemoryStream(StringToByteArray("0300002902f0803203000000010002001400000401ff040080"));
-            var t = await TPKT.ReadAsync(m);
+            var t = await TPKT.ReadAsync(m, CancellationToken.None);
          }
 
         [TestMethod]
@@ -77,7 +79,7 @@ namespace S7.Net.UnitTest
             var t = COTP.TSDU.Read(m);
             Assert.IsTrue(expected.SequenceEqual(t));
             m.Position = 0;
-            t = COTP.TSDU.ReadAsync(m).Result;
+            t = COTP.TSDU.ReadAsync(m, CancellationToken.None).Result;
             Assert.IsTrue(expected.SequenceEqual(t));
         }
 
--- a/S7.Net.UnitTest/ProtocolTests.cs
+++ b/S7.Net.UnitTest/ProtocolTests.cs
@@ -35,6 +35,33 @@ namespace S7.Net.UnitTest
         }
 
         [TestMethod]
+        public async Task TPKT_ReadDelayedAsync()
+        {
+            var fullMessage = ProtocolUnitTest.StringToByteArray("0300002902f0803203000000010002001400000401ff0400807710000100000103000000033f8ccccd");
+            var m = new MemoryStream();
+            m.Write(fullMessage, 0, 2);
+            var tcs = new TaskCompletionSource<bool>();
+            tcs.Task.ContinueWith(x => m.Write(fullMessage, 2, fullMessage.Length - 2));
+            var t = TPKT.ReadAsync(m);
+            tcs.TrySetResult(true);
+            await t;
+        }
+
+        [TestMethod]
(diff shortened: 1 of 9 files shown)


scamille:
One thing I still don't understand is why with just your initial ReadFixed implementation this would result in a infinite blocking compared to the current code.

With the current code, any read/write blocks will time out based on what is set in https://github.com/S7NetPlus/s7netplus/blob/develop/S7.Net/PLC.cs#L210 to the Receive and Send Timeout.

With the ReadFixed implementation, you might in the worst case wait longer if you get your whole message 1 byte after another. As soon as you get 0 bytes back (EOF), you bail as well.

Since even with the current implementation, the first read call could block indefinitely instead of giving you less bytes than requested, this can block indefinitely as well. We might want to change the default at https://github.com/S7NetPlus/s7netplus/blob/develop/S7.Net/PLC.cs#L21 to something more sensible, but that is another topic.

Or am I missing something?

jakob-ledermann:
Focusing on the non async implementation, the current implementation does a special casing of Networkstream and a generic Stream implementation, that can not block as I don't know about any Stream implementation, that CanSeek and blocks on read, any other stream must support timeouts and have a timeout set earlier than never.

The special case for NetworkStream checks whether data are available and only reads when data is available (it is much closer to the behaviour on nonblocking linux sockets) and then blocks the current thread for a short amount of time (originally Thread.Sleep, but replaced with SpinWait as it was the first way to block the current thread supported on .Net Standard 1.3) the potentially blocking code is only entered, when it can not block. Given there are no two threads or tasks reading at the same time, which is a bad idea for different reasons.

As I somewhere in the documentation a remark the connection should be closed after a readtimeout, I tried to avoid that situation, in addition to avoiding changing the default in https://github.com/S7NetPlus/s7netplus/blob/develop/S7.Net/PLC.cs#L21 as I wasn't sure whether this has additional side effects, I am missing.

scamille:
I took your first two commits, and adjusted some things, and in particular the Unit Test. I added a fake stream giving 1 byte at a time. This breaks the previous S7Net code without the ReadFixed extensions, without introducing any delay based special cases for now.
See https://github.com/S7NetPlus/s7netplus/compare/develop...scamille:fb-readFixed

What do you think?

jakob-ledermann:
I think the name ReadExact is a better name for ReadFixed (I read it in the [rust documentations](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact) and liked it)
If the default ReadTimeout is changed I think that would be a workable solution.

I'm not completly happy with the special casing of the networkstream in my current implementation.
The current set of unit tests does not catch the case where the answer is malformed (reports a different length than the sent length) and I would prefer the message to fail in this case vs. endlessly blocking the users code or even having two active reads on the connection active at the same time.

scamille:
My plan is to investigate what default timeouts are set for a connection and maybe we can use those. I would assume that .NET or your OS makes a sensible choice for those.

If we make the custom timeout variables in the PLC object nullable this could represent this "default" state.

And yes ReadExact sounds good, I just did not change it. My commit was just intended as a quick hack and would need cleanup anyway.

scamille:
> The current set of unit tests does not catch the case where the answer is malformed (reports a different length than the sent length) and I would prefer the message to fail in this case vs. endlessly blocking the users code or even having two active reads on the connection active at the same time.

What exactly do you mean with this?

I think we can safely assume that a call to ```var byteRead = Stream.Read(buffer, offset, count);``` will have put exactly bytesRead number of bytes into the buffer starting from position offset. Otherwise quite a lot is broken with any .NET Stream implementation.

I you are talking about the 2-phase TPKT Reads: Yes it could be that the first packet receives a malformed length (eg. 65k) and you will then try to get that amount of bytes from the stream.
But unless you have a malicious attacker only giving you 1 byte at a time and thus increasing your total timeout time for the whole request, this should make no difference to the current implementation. Once you get a Read Timeout you also will bail out of ReadExact through an exception.
And if you have no Read Timeout at all, you can already block on the very first Read call forever.

So my main point is: Adding ReadExact likely does not make the timeout behaviour of the library worse compared to the current code in any realistic way.

And since this should definitely fix problems people have when receiving partial responses and getting Exceptions, I think we should resolve this with just those changes for now.

It might certainly be worth thinking about having timeout and async cancellation options on all public PLC functions, providing a fixed and reliable timeout for the complete operation. But again, different topic.

jakob-ledermann:
> So my main point is: Adding ReadExact likely does not make the timeout behaviour of the library worse compared to the current code in any realistic way.

This is actually a valid point.

And yes I was talking about malicious (or erroneous) TPKT datagrams. There is a second bigger issue, which is also in my opinion of topic for now: I managed to send a second request to the plc before the first one was answered by the plc.

So I agree with your PR variant #291 for now and set a usefull readtimeout in my use case.
Unfortunatly DataAvailable is not part of the standard Stream-API so my speciallisation on NetworkStream can not be applied generically and also not tested easily.

scamille:
As far as I understand it, the S7 protocol would allow sending a "identifier" with each request, and you can expect to get that identifier back in the response from the PLC.

S7Net.Plus just always sends the same identifier (0 I believe) and does not check anything. We could check that, further increasing robustness.

But one fundamental limitation of the library certainly is that for the single socket we have per PLC, there can only be 1 request active during any time. Having multiple requests would require a fundemantally different architecture, and make things way way too complicated.

The appropriate solution for S7.NetPlus would be to either add some locks or at least be informative about it being neither multi-thread-safe nor truly suitable for unlocked async operations.

I personally have async locks available for the software where I integrated S7.NetPlus library, so did not encounter any of the problems. But I expect that specifically for users of the async api, this could lead to some confusion.

PR #291, opened 2020-08-19
Fixed size reads version 2

scamille:
Alternative resolution for #289

based on the initial work of @jakob-lenderman with some small adjustments. It tries to resolve the initial problem of terminating Stream.Reads too early if the call returns less bytes than requested.

Adds a Unit-Test with a Test Stream class instead of MemoryStream to verify that this should be working.


mycroes (approved):
LGTM, if you rebase I'll merge this as well.

scamille:
Ok done.

Issue #171, opened 2018-07-30
Read StringEx fails if DB contains rubish


bauerju:
Today I had an exception while reading a String from the PLC. Someone wrote to the wrong DB. This destroyed the used string length information.   
Example: Tag was declared as DB1.DBB0 with DataType StringEx and Length of 4 chars. This takes 6 Bytes in plc. Something like that: 04 04 31 32 33 34. 

Someone wrote to DB1.DBB1 a value higher than  4. 
In the DB was something like that: 04 05 31 32 33 34. 

The read came back with 6 bytes. But the conversion to String failed in StringEx.From,ByteArray(Byte[] bytes).
Yes I know. This is something which should not happen.  
I changed the source in  S7.Net.Types for me this way:

        public static string FromByteArray(byte[] bytes)
        {
            if (bytes.Length < 2) return "";

            int size = bytes[0];
            int length = bytes[1];

            // Check if byte array is long enough
            if (bytes.Length - 2 < length)
            {
                length = bytes.Length - 2;
            }

            return System.Text.Encoding.ASCII.GetString(bytes, 2, length);
        }

Jürgen

mycroes:
Hi @bauerju,

Honesty I think an exception would be in place here, since the data is invalid. Masking that by returning the garbage as string or returning an empty string isn't always a good idea, although I can image we could create a `bool TryConvertFromByteArray(byte[] input, out string value)` on `StringEx` to allow for the behavior you implemented (although I would return `null` when `length < 2`). I see that the current exception has the following message: **Index and count must refer to a location within the buffer.**, wrapping this in another exception with a message that the string conversion failed would seem to address this issue if you ask me.

So in short, my proposal:
1. Wrap the exception with a concise message
2. If that doesn't float your boat, add `bool StringEx.TryConvertFromByteArray(byte[] bytes, out string value)`

mycroes:
To amend my previous message, I just noted that the current code returns `""` on `bytes.Length < 2`, so ignore that comment.

bauerju:
Hi @mycroes 
Yes, you are right. An exception should be fired. But only if this is a single read. Just one item.
In the case of a read with multiple items, my opinion is that the dataItem should better get an error flag not to break the whole read. 

Jürgen

mycroes:
Hi @bauerju,

I agree about not breaking the rest of the read, so what I'll implement is to set all the fields that can be set and to throw an `AggregateException` to combine the individual errors at the end of the read action.

[...]

PR #292, opened 2020-08-19
Adjust StringEx ToByteArray.

scamille:
- Do not allow null string to be passed, raise ArgumentNullException.
- Do not allow string whose ASCII representation is longer than the reserved length, since this currently leads to silent data loss.
- Always write the full binary data length of 2 + reservedLength, since that is what the binary representation of that string is in S7 memory, even if some tail bytes are unused by the current string.

I also suspect that S7WriteMultiple would have chocked on that last bit, but I am not sure. There aren't any tests for writing multiple Dataitems right now.

Adjust tests accordingly. Mostly add some tail bytes where necessary, and assert on exceptions where this is now required.

### DIFF
--- a/S7.Net.UnitTest/TypeTests/StringExTests.cs
+++ b/S7.Net.UnitTest/TypeTests/StringExTests.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 using S7.Net.Types;
+using System;
 
 namespace S7.Net.UnitTest.TypeTests
 {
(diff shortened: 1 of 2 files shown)


mycroes (approved):

mycroes:
As with all the other PR's I think this is a useful change. I have literally next to none PLC experience, so I once chose to only write the affected bytes, but from recent experience I would've probably made the same choice as implemented here.

mycroes:
This also needs a rebase BTW.

scamille:
As mentioned in #171 it might be a sensible addition to offer a API of reading n variables and not failing hard if just some of them fail the conversion. It would instead either offer a status flag or on-demand conversion.

But I think that is only partially related to checking data conversion very strictly by default. For ReadStruct and others, we certainly do not want partial results in the object you get from the library. Either everything converted properly or you get an exception.

scamille:
Rebased

PR #293, opened 2020-08-19
Add nullable support throughout the whole library.

scamille:
This requires reference types that can be null to be annotated by a ? operator, similar to value types.

This gives the advantage that the compiler can warn against any nullreference exceptions, of which this commits elimits a few.

To make the underlying protocol implementation not any more complicated and to eliminate existing problems, and to give more precise error reporting, I replaced some return null statements with explicit Exceptions. This lead to the assumption that those core protocol functions always return non-null objects if they do not throw, making the PLC code simpler.

Adjust some NotConnected tests to look for explicit PlcException instead of NullReferenceException.

### DIFF
--- a/S7.Net.UnitTest/S7NetTestsAsync.cs
+++ b/S7.Net.UnitTest/S7NetTestsAsync.cs
@@ -591,15 +591,13 @@ namespace S7.Net.UnitTest
 
 
         [TestMethod]
-        [ExpectedException(typeof(NullReferenceException))]
         public async Task Test_Async_ReadBytesReturnsNullIfPlcIsNotConnected()
         {
             using (var notConnectedPlc = new Plc(CpuType.S7300, "255.255.255.255", 0, 0))
             {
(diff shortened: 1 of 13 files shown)


mycroes:
Hi @scamille,

Yes this was motivation for #296, which is superseded by #302. Can you update your PR and see if it works now? I will definitely accept this PR; the only reason I personally see to avoid nullable reference types is legacy codebases.

scamille:
Adjusting S7NetPlus to have nullable annotations should not affect any user of the library, unless nullable support is enabled there as well. It is after all just a compiler-internal annotation, and does not change anything about the fact that reference type variable can still be null.

mycroes:
HI @scamille, I understand how the feature works, I just meant that I wouldn't enable nullable reference types on a 10k+ lines codebase if there's no obvious reason to do so (or at least professionally, since I can't substantiate the investment with regards to the benefits). Nevertheless, I'm happy that you made this change, merging right away.

PR #297, opened 2020-08-21
Change test project to a fixed win-x64 runtime identifier and upgrade…

scamille:
… Snap7

Downlaoded Snap7 1.4.2 64bit DLL from https://sourceforge.net/projects/snap7/files/


mycroes:
Hi @scamille, please rebase this one as well. It seems to build/test just fine on AppVeyor as well, so I don't see much reason not to merge this.

scamille:
I think the original motivation was that this might also reduce sporadic Unit Test failures, even though I have no clue if it does make a difference.

scamille:
Looks like this might really help with sporadic tests failing in CI, in particular T33_WriteLargeByteArray

scamille:
Going to retract this, as #306 contains both the upgraded Snap7 library as well as some other, better adjustments than this.

PR #298, opened 2020-08-21
Add LReal support, kill old "Double" sham.

scamille:
- Adds true support for 64bit double / LReal datatype.
- Set old Types.Single and Types.Double to obselete. Both class names use .NET types instead of S7 type names, contrary to all other types.
- Remove already obsoleted conversion from DWord to Real. Why is this even necessary?
  For users caring about converting from DWord, they can still convert to single. But unless we get LWord support, there won't be a direct conversion to double/LReal.
- Adjust unit tests by removing rounding, testing directly double read/writes.

There is quite a bit of breaking changes at least in the automated Struct and object functions which automatically translate .NET types to appropriate S7 types.
My consideration was that if we ever want to support 64bit types, there is no way against breaking those existing incorrect conversions from 64bit .NET double to 32 bit S7 Real variables.

### DIFF
--- a/S7.Net.UnitTest/S7NetTestsAsync.cs
+++ b/S7.Net.UnitTest/S7NetTestsAsync.cs
@@ -876,17 +876,6 @@ namespace S7.Net.UnitTest
             Assert.AreEqual(tc.CustomTypes[1].Bools[1], tc2.CustomTypes[1].Bools[1]);
         }
 
-        //[TestMethod]
-        //public async Task Test_Async_ReadWriteDouble()
-        //{
-        //    double test_value = 55.66;
-        //    await plc.WriteAsync("DB1.DBD0", test_value);
-        //    var helper = await plc.ReadAsync("DB1.DBD0");
-        //    double test_value2 = Conversion.ConvertToDouble((uint)helper);
-
-        //    Assert.AreEqual(test_value, test_value2, 0.01, "Compare Write/Read"); //Need delta here because S7 only has 32 bit reals
-        //}
-
         [TestMethod]
         public async Task Test_Async_ReadWriteSingle()
         {
--- a/S7.Net.UnitTest/S7NetTestsAsync.cs
+++ b/S7.Net.UnitTest/S7NetTestsAsync.cs
@@ -101,19 +101,12 @@ namespace S7.Net.UnitTest
 
         /// <summary>
         /// Read/Write a single REAL with a single request.
-        /// Test that writing a double and reading it gives the correct value.
+        /// Test that writing a float and reading it gives the correct value.
         /// </summary>
         [TestMethod]
         public async Task Test_Async_WriteAndReadRealVariables()
         {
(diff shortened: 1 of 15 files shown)


mycroes:
Will be happy to accept. I might have mentioned this before, but I'm not the (original) author of this library. I favor standardization as performed here, just as well as deleting invalid / unnecessary options like DWord to Real conversion.

It can sometimes be hard to justify these changes to end users, but it helps to have allies 😉

S7.Net/Protocol/Serialization.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
mycroes, line 3: Accidental inclusion?

S7.Net.UnitTest/S7NetTestsAsync.cs
```diff
@@ -883,16 +876,16 @@ public async Task Test_Async_ReadClassWithArrayAndCustomType()
...
         }
 
-        [TestMethod]
-        public async Task Test_Async_ReadWriteDouble()
-        {
-            double test_value = 55.66;
-            await plc.WriteAsync("DB1.DBD0", test_value);
-            var helper = await plc.ReadAsync("DB1.DBD0");
-            double test_value2 = Conversion.ConvertToDouble((uint)helper);
+        //[TestMethod]
```
mycroes, line 879: Let's delete this, git will remember anyway.

S7.Net/Types/Class.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
mycroes, line 3: Accidental inclusion?

mycroes (changes requested):

S7.Net.UnitTest/S7NetTestsAsync.cs
scamille: I probably thought about adding long-word support to bring this test back. But I don't really see any value in word to other type conversions anyway, so can just remove it completely.

S7.Net/Protocol/Serialization.cs
scamille: Yes. Fixed in 6640a39

S7.Net/Types/Class.cs
scamille: Yes. Fixed in 6640a39

mycroes (approved):

mycroes:
Can you rebase this as well? If there's any specific order that you want the open PR's merged in (I'm betting on more rebases required) let me know.

scamille:
let's just do this one first since you reviewed it :) Next the PDU checks I guess, then Unit Test Snap7 upgrade. The string changes need some discussion maybe, dunno how much it will break things.

mycroes:
Thanks again @scamille!

PR #304, opened 2020-09-04
Add github actions workflow

scamille:
This will need quite a bit of testing until it is ready, which can only be done in here :-)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,8 +1,7 @@
 name: CI
 
 on:
-  pull_request
-  push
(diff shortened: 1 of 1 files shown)


scamille:
Seems this does not work out of the box, which makes sense since you don't want to run arbitrary workflows from a pull request.

The motivation behind this:
- I am learning/playing with GitHub Actions anyway right now.
- Sporadic Unit Test Failures on AppVeyor
- Faster than AppVeyor
- Can be extended to other OS than windows.

PR #306, opened 2020-09-06
Add Github Actions workflow

scamille:
- Adds github CI pipeline for testing as well as packaging (as the existing appveyor pipeline does)
- Includes test run for ubuntu-latest on netcoreapp3.1
- Includes using 64bit Snap7 dll version 1.4.2
- Adjust Snap7 Test Server Network port to a value > 1000 so tests can run on linux without elevated privileges.
- Fix default PLC TcpClient Timeout values to 0 to fix runtime errors on linux. This is the value specified in the .NET API and should do the same as the existing -1 value on windows, namely no timeout (infinite).
- Adjust Snap7 .NET wrapper to use platform independent dll name.

Example Run:
https://github.com/scamille/s7netplus/actions/runs/241585909


mycroes:
Definitely interested in this, however I want to do some additional research/testing before I merge this. I also deploy from AppVeyor (manual deploy from build for version tag), so that's at least one thing I also need when using GitHub actions.

scamille:
They should definitely run in parallel to begin with, so that it can be verified to work reliably.

I tried to add artifact creation to this similar to how you set it up for appveyor. But I don't know how exactly your system works when doing releases. that something which would need some refinement if you want it to work from it.

Or you just keep this as an additional CI Test only and continue using appveyor for deployment indefinitely.

mycroes:
Been reading a bit on workflows, so I at least figured the following:

- Publishing to NuGet shouldn't be too hard
- There's `on: release: types: [created]' which runs when a release is created, this could perhaps be the right time to publish the NuGet package. Not sure if this works with automatically created releases on version tag creation, which I currently use.
- The workflow created here is quite complex to achieve build+test, even though it does include all the versioning stuff from the AppVeyor setup.

So I'm wondering, I think we should bring this back to a `build+test` workflow (instead of `main`) which doesn't bother with all the versioning stuff, and a release/publish workflow which publishes the package. I'm not asking you to create this, if you want to adapt this to build+test I'll merge this and when I have some more time to figure out the deployment to NuGet I can add that myself as well.

scamille:
Ok so if I understand you correctly, bring this back to only build + test without GitVersion, Pack and Upload Artifacts which bloats it so much.

how should thee file be named if not main.yml?

The name of the workflow is already 'CI'.

Might make sense to have a suitable name which is the same for both the workflow file and name.

scamille:
As far as I understand, the main reason to keep things inside the same workflow is so that jobs can depend on each other. It might make sense to only run deploy after build + test was successful.

On the other hand, triggers are only defined once per workflow. Thus it would make sense to have a build + test workflow for push/pull request, and the deploy workflow for tags created.

I guess you can somehow assume that a tag will always be created only for a commit for which the build + test pipeline was successful, otherwise it wouldn't have been merged into master in the first place.

mycroes:
> Ok so if I understand you correctly, bring this back to only build + test without GitVersion, Pack and Upload Artifacts which bloats it so much.
Yes!

> how should thee file be named if not main.yml?
build+test.yml?

> The name of the workflow is already 'CI'.
I'm not sure if CI is technically correct (which depends on how one would define CI, because the fact that a PR builds doesn't mean it'll also get merged).

> As far as I understand, the main reason to keep things inside the same workflow is so that jobs can depend on each other. It might make sense to only run deploy after build + test was successful.
Yes, I guess it should also be part of the deploy workflow.

> On the other hand, triggers are only defined once per workflow. Thus it would make sense to have a build + test workflow for push/pull request, and the deploy workflow for tags created.
Exactly.

> I guess you can somehow assume that a tag will always be created only for a commit for which the build + test pipeline was successful, otherwise it wouldn't have been merged into master in the first place.
master is actually a protected branch (and develop as well I think), so even I can't push directly to it right now (of course I could unprotect them, but you get the point), which means you're correct about the verification always happening upfront.

scamille:
Ok so got it separated between Test and Deploy (which has no active job at the moment; just wanted to store the existing work on importing GitVersion, Pack and Upload Artifacts).

This whole Pipeline stuff is a lot of trial and error, and my scripting skills are not the best :)

Anyway, please have a look. Test run: https://github.com/scamille/s7netplus/actions/runs/246964386

scamille:
Replaced by #319 so that #318 can be merged independently from it.

PR #307, opened 2020-09-06
Add CancellationToken support to async functions.

scamille:


scamille:
@mycroes Even though this is not fully finished, feel free to take a look at it and give me feedback if you think something fundamental should be done differently.

My current goal was to just make sure that all public async functions get an optional CancellationToken parameter at the end, and that it is passed through to all Stream/TcpClient functions which offer a CancellationToken parameter.

I don't think it is worth doing many CancellationToken checks / ThrowIfCancelled calls ourself. In a normal network based usage scenario you probably have 1-5% of the time spent on serializing as well as doing the protocol communication, and the rest being very slow network communication.

I also don't have any plans on offering something other than cooperative cancellation through the stream .NET library functions, at least not as a baseline.
Even adding something to Open() is a bit fishy, since it will only affect the communication after the connection is established.

mycroes:
Nothing fundamentally wrong with your code, but what's fundamentally wrong with the library is that's there's no seperation of concerns between building messages, parsing messages and doing network communication. This also means that the code is riddled with `stream.Write`(`Async`) and `stream.Read`(`Async`) calls. I definitely want to add closing of the socket on cancel, but I don't think the order of implementing that matters much.

I've also been thinking about removing the sync implementation and just forwarding all the calls to the async implementations.  OTOH with proper SoC it's probably peanuts to keep the sync implementation with calls to sync `Read` and `Write`.

scamille:
Yes I have been thinking about reducing the amount of code as well.
One thing I want to do is to move all those serialization stuff into extension methods, so the PLC class can focus on actually reading/writing binary data (which is either a single byte array or a list of DataItems).

I don't know if its a good idea to have all synchronous functions just wrap the async ones. It certainly would reduce the maintenance burden quite a lot. Guess I need to read up on this a bit.

Yes it would definitely help to separate concern of building the message, parsing them and doing the network communication more clearly, or at least make sure nothing is duplicated. I tried to do some of that in the previous fb-var pull request, so it should at least be a bit better.

it might also be possible to merge the ReadMultipleVars with the ReadBytes function, since the latter is a special case of the former. But it is a bit tricky since the former does not support operations larger than Max PDU.

scamille:
https://devblogs.microsoft.com/pfxteam/should-i-expose-synchronous-wrappers-for-asynchronous-methods/ has some arguments against wrapping async inside sync calls. There are good reasons to avoid it, but it might still be the sensible thing to do for us, assuming we take care of some of the pitfalls.
Got There from https://stackoverflow.com/questions/19865823/how-can-i-implement-both-an-async-method-and-its-synchronous-counterpart / https://stackoverflow.com/questions/14870478/write-a-well-designed-async-non-async-api/14876375#14876375

mycroes:
I'm actually only offering async in [Sally7](/mycroes/Sally7), but yes, there's an existing userbase here that might have issues with that. That's also one of the reasons that this project needs a proper sample implementation (to educate users).

About the wrapping: network connections are inherently async, so acting like we're doing a sync call because TcpClient implements a sync API doesn't really differ much from moving the async-to-sync barrier one level up. I don't know if there's anything in the sync methods that's different than blocking till the result is available, because that would be the only reason for using the sync TcpClient API from my point of view (and only for backwards compatibility in the sync methods, because the right approach would be to use the async API anyway).

scamille:
Well it would definitely simplify a lot of things if we can get rid of sync implementations, so I am all in on it.

My plan would be to first complete this PR.
Then either
- Yank serialization stuff into extension methods
- replace Sync calls with async wrappers

And after that we can look some more into cleaning up the internals of async connect/read/write, better separate the different steps.

mycroes:
One minor comment, I don't think extension methods would offer much of an advantage, while OTOH they would represent the cor functionality of this library, which makes it questionable whether it should only be accessible through extension methods anyway. Other than that, I think your plan is OK. Will probably look into merging this PR this evening.

mycroes (approved):

Issue #275, opened 2020-07-14
How to handle the PLC disconnection error?


nayanavv:
How to handle the error thrown when the PLC (ethernet port) is disconnected from the windows application. How to handle this error?  By that I meant how to not suspend the thread and keep the application running.
Also I'm getting different error for ethernet disconnection and PLC power off, how is that possible?

Plc is not getting disconnected when PLC Close function is called in the form.closed event.

>  PLC.Close()

Please help with any possible solutions

kivannc:
I have a similar problem. In some computers when the is connection is gone for a second. IsConnected stays true but it always gets read error and shows wrong values. For now I have counter for errors. If it continues for a while. I reinitialize the connection.

nayanavv:
> I have a similar problem. In some computers when the is connection is gone for a second. IsConnected stays true but it always gets read error and shows wrong values. For now I have counter for errors. If it continues for a while. I reinitialize the connection.

Hi kivannc,
Could you please explain how you did this? Reinitialize means do I have to restart the application?

kivannc:
I am just Closing and Openning Connection again. Instance.Close(); Instance.Open();

mycroes:
This might also be affected by #349.

Issue #252, opened 2020-01-29
Unable to read PLC DB data


05749:
Hi, I am using the S7-1200 and TIA portal V1.4.
I am running Readbytes from the S7.net library, but I get the following error:
 "Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host."
I can do plc.open () and plc.close () without any problems.
Is there any solution? Any advice is welcome.
thank you in advance.
![plc program](https://user-images.githubusercontent.com/60371927/73341639-6393b500-42c0-11ea-96a3-a3b7a6bcc725.png)

FalcoGoodbody:
Did you check that the DB is not optimized?

05749:
Thank you for your comment.
Yes, I unchecked DB optimization according to the documentation.
![image](https://user-images.githubusercontent.com/60371927/73581158-a8d80280-44cb-11ea-8599-6a426cbd7ad2.png)
Do I need other settings on the PLC besides the contents of this document?
Thank you.

allatrue:
Did you enabled PUT/GET communication for the CPU?

05749:
Yes, I have enabled PUT / GET communication for the CPU .

mycroes:
Try to remove the Plc.Isconnected check, it's broken.

05749:
I thought Plc.Isconnected was a way to verify communication with the PLC, but does it affect Plc.ReadBytes?

mycroes:
I believe it accidentally terminates the connection. I'm way behind on fixing issues like this, because obviously this shouldn't be the case. Let me know if it works without the IsConnected check.

05749:
Thank you for your comment.
I was able to read the DB by removing the IsConnected check.
![Improved PLC program](https://user-images.githubusercontent.com/60371927/73617752-65190080-4665-11ea-952a-20f2dcf794a6.png)
thank you very much.

FalcoGoodbody:
Well...that´s strange.
I am using the PLC.IsConnected to read data from a S7-1500 at there it works fine so far.
Maybe it is problem depending to the PLC used?

mycroes:
Maybe one of you isn't using the latest version?

FalcoGoodbody:
I´m using Version 0.4.0 from the last year. Via NuGet there is no newer one available.
For me everything works fine and I am absolutely happy with this library.

05749:
I using the same S7.Net version 0.4.0.
[...]

Issue #208, opened 2019-02-20
I have problems connecting to an S7 200


frustis:
My code:

`using System;
using System.Collections.Generic;
using System.Linq;
using System.Threading.Tasks;
using System.Windows.Forms;
using S7.Net;

namespace WindowsFormsApp1
{
    static class Program
    {
        /// <summary>
        /// Punto de entrada principal para la aplicación.
        /// </summary>
        [STAThread]
        static void Main()
        {
            // Application.EnableVisualStyles();
            Plc plc = new Plc(CpuType.S7200, 10.127.85.210, 0, 2);
            /// plc.
            plc.Open();

            Console.WriteLine(plc.IsAvailable);
            Console.WriteLine(plc.IsConnected);
            // var dwords = plc.Read(DataType.DataBlock, 1, 0, VarType.DWord, 20);
            // ushort result = (ushort)plc.Read("DB1.DBW0");

        }
    }
}`

In the plc.Open () I get an exception:

Excepción producida: 'S7.Net.PlcException' en S7.Net.dll
Excepción no controlada del tipo 'S7.Net.PlcException' en S7.Net.dll
Couldn't establish the connection to 10.127.85.210.
Message: TPKT is incomplete / invalid

Any Idea?

Thanks

mycroes:
Hi @frustis,

At first I saw the PLC type in your code didn't match the issue title, but since you corrected that I'm assuming you are using the right values now. Unfortunately your issue is hard to debug. The PLC refuses the connection, the reason is unclear. I'm guessing it has to do with some data in the Connection Request as sent by S7NetPlus, but what part I don't know. There are more open issues related to this, I'm trying to improve the situation but I don't have much time these days... I will try to get back on this soon. In the meantime, have you tried any other library for S7 communication to see if it works?

mycroes:
Just spotted the `plc.IsConnected` usage, unfortunately that kills the connection. Please retest this without `plc.IsConnected` and preferably against a newer S7NetPlus version.

scamille:
IsConnected does not kill your connection I believe, only IsAvailable might.

The main problem here should be fixed in version 0.7.0, so just try again with the latest version please.

mycroes:
Also, this would probably be fixed by #396 when you specify TSAP's manually, because from what I understood the S7 200 doesn't support a fixed set of TSAP addresses as most newer PLC's do...

mycroes:
Closing, feel free to reopen if the issue persists.

Issue #154, opened 2018-07-06
IsConnected don't check if connection setup is done (Open())


rapha-dev:
As title says, running a `Read` (or any read/write action) without prior `Open` will throw a `TPKTInvalidException`.
`IsConnected` should check actual connection by sending a empty PDU (as mentioned in source already).

This behavior breaks old implementations where `IsConnected` actually checks for prior `Open` (before 0.1.8).

mycroes:
The IsConnected property of `PLC` now reflects the `IsConnected` state of the underlying `TcpClient`. #349 will actually add to that by having the connection closed on errors.

mycroes:
Also see #523, I tink this can be closed now.

PR #311, opened 2020-09-08
Remove IsAvailable, document IsConnected

mycroes:
Addresses #275, #252, #208, #154.

### DIFF
--- a/S7.Net.UnitTest/S7NetTestsSync.cs
+++ b/S7.Net.UnitTest/S7NetTestsSync.cs
@@ -933,7 +933,14 @@ namespace S7.Net.UnitTest
             S7TestServer.Stop();
 
             var unreachablePlc = new Plc(CpuType.S7300, "255.255.255.255", 0, 2);
-            Assert.IsFalse(unreachablePlc.IsAvailable);
+            try
+            {
+                unreachablePlc.Open();
+            }
+            catch
+            {
+            }
+            Assert.IsFalse(unreachablePlc.IsConnected);
         }
 
         [TestMethod]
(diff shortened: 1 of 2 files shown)


scamille (commented):
At first I thought it might be a good idea to just let IsAvailable call IsConnected first, and only if it is not try to connect to the PLC.

But on the other hand, IsAvailable is just a property and tried to do too much, hiding large function calls in it.

So I think it is a sensible choice to just force users to do the Open() call themselves, with all the exception handling necessary for it.
It might even be worth deprecating IsAvailable and pointing to IsConnected.

mycroes:
Hi @scamille, Thanks for taking a peek at this. I was actually slightly confused (as you noticed from my comments with regards to IsConnected on issues) between IsAvailable and IsConnected. I was also thinking about removing IsAvailable altogether. Either way, I was short on time yesterday, so I haven't made up my mind yet. 😄

scamille:
I do like the comments, it really highlights the problems with relying on IsConnected giving the full truth about the connection status.
Even if it would send a keepalive/other dummy data, it would still only give you the IsConnected status at that point of time. A Read/Write after that could still run into a unconnected connection and needs to handle it somehow gracefully. 

You should be able to reference the IsConnected XML documentation from IsAvailable with /// <inheritdoc cref="Plc.IsConnected"/>

And yes adding [Deprecated] would be appropriate, maybe with some words about calling Open() yourself. We certainly don't need IsConnected twice forever.

mycroes:
The comment actually comes from TcpClient.Connected. I think I might just remove IsAvailable altogether, this breaks it anyway for people relying on it to connect this way...

PR #314, opened 2020-09-12
Cleanup OpenAsync function.

scamille:
- Separate into Connect and EstablishConnection step.
- Remove redundant null checks for returned data.
-  Only assing PLC stream object once we fully established a connection, and Close otherwise.
- Replace sync implementation with Async call.


scamille:
Should I add .ConfigureAwait(false) to all internal async call?

I haven't really done that much library work. From what I have read it should help by not always capturing synchronization contexts, which might help when used from GUI applications.

mycroes:
> Should I add .ConfigureAwait(false) to all internal async call?
> 
> I haven't really done that much library work. From what I have read it should help by not always capturing synchronization contexts, which might help when used from GUI applications.

I'd say a definite yes, because it's not something you can fix while consuming a library. If you need to return to the same synchronization context as consumer, that'll still happen where you're actually awaiting the library calls.

S7.Net/PlcAsynchronous.cs
```diff
@@ -54,15 +78,12 @@ public async Task OpenAsync(CancellationToken cancellationToken = default)
...
-            MaxPDUSize = (short)(s7data[18] * 256 + s7data[19]);
-        }
-
-        private async Task ConnectAsync()
-        {
-            tcpClient = new TcpClient();
-            ConfigureConnection();
-            await tcpClient.ConnectAsync(IP, Port);
-            _stream = tcpClient.GetStream();
+            // TODO: check if this should not rather be UInt16.
```
mycroes, line 81: No, it's a byte; see [RFC905](https://tools.ietf.org/html/rfc905), section 13.3.4 b.

S7.Net/PlcAsynchronous.cs
```diff
@@ -54,15 +78,12 @@ public async Task OpenAsync(CancellationToken cancellationToken = default)
...
+            MaxPDUSize = Types.Int.FromByteArray(s7data, 18);
```
mycroes, line 82: Which also makes this invalid (but possibly the next byte is always 0, or at least always 0 in response to _our_ ConnectionRequest).

S7.Net/PlcAsynchronous.cs
```diff
@@ -54,15 +78,12 @@ public async Task OpenAsync(CancellationToken cancellationToken = default)
...
+            if (MaxPDUSize <= 0 )
```
mycroes, line 83: [RFC905](https://tools.ietf.org/html/rfc905) 6.5.4 j status that the minimum is actually 128, which also is the maximum length for the CR TPDU. Also, this number is unsigned and would logically always be unsigned in any protocol, so the < 0 part would most logically be invalid interpretation on our side 😜.

S7.Net/PLC.cs
```diff
@@ -85,7 +85,7 @@ public bool IsAvailable
             {
                 try
                 {
-                    Connect();
+                    ConnectAsync().Wait();
```
mycroes, line 88: Use `ConnectAsync().GetAwaiter().GetResult()` instead. I don't know all the details, but for one it doesn't wrap any exception in an `AggregateException`, which is already enough to warrant the change.

S7.Net/PlcSynchronous.cs
```diff
@@ -17,34 +17,9 @@ public partial class Plc
...
-
-                //Check for S7 Ack Data
-                if (s7data[1] != 0x03)
-                    throw new InvalidDataException("Error reading Communication Setup response", s7data, 1, 0x03);
-
-                if (s7data.Length < 20)
-                    throw new WrongNumberOfBytesException("Not enough data received in response to Communication Setup");
-
-                MaxPDUSize = (short)(s7data[18] * 256 + s7data[19]);
+                OpenAsync().Wait();
```
mycroes, line 22: Use `OpenAsync().GetAwaiter().GetResult()` instead.

S7.Net/Types/Int.cs
```diff
@@ -10,15 +10,15 @@ public static class Int
...
-            if (bytes.Length != 2)
+            if (bytes.Length < offset + 2)
             {
-                throw new ArgumentException("Wrong number of bytes. Bytes array must contain 2 bytes.");
+                throw new ArgumentException($"Wrong number of bytes. Bytes array must contain at least {offset+2} bytes.");
             }
             // bytes[0] -> HighByte
             // bytes[1] -> LowByte
-            return (short)((int)(bytes[1]) | ((int)(bytes[0]) << 8));
+            return (short)(bytes[offset + 1] | (bytes[offset + 0] << 8));
```
mycroes, line 21: Not a fan of this change. It make Types.Int different than all the other types. In Sally7 I use `Span` because you can get ranges without copies, in the long run that might be a better solution here as well.

scamille:
Ok I will add it for all relevant OpenAsync calls then, so that we can get this done in here.

mycroes (changes requested):
I wrote Sally7 by reading whatever documentation was available, so I actually knew where to look for some of the changes in here.

S7.Net/PlcAsynchronous.cs
scamille: Ok so this doesn't really make sense to me. We negotiate high PDU sizes > 255 with Snap7 or modern S7 PLCs. How would that be possible if this is only 8bit?

The Signed Part really did not make sense to me / I was sceptical about, since those are quite rarely used in such low level protocols and especially not for a size type.

Of course using signed here doe not matter unless you get a value above 2**15, but it would still be better to just handle it correctly.

S7.Net/Types/Int.cs
scamille: Yes I agree with your critique. The idea was to reuse the already available big endian serialization we had. But I guess for such a one-liner we can just use hand-rolled conversion as before.

S7.Net/PlcAsynchronous.cs
scamille: Also the MaxPDUSize bytes are at address 18 of the PDU Payload and that is usually deep into the S7 protocol, at least for the other functions. So I am not really sure if this matches your reference in the RFC document. But I am really not good at reading this thing.

S7.Net/Types/Int.cs
scamille: Changed back in 4a72c35

scamille: Yes Span is nice and all but I don't see that happening here for a long time.
I also did some test with the new Memory class and spans here in S7.NetPlus, and while it might certainly save some cpu cycles, the overall execution time difference is not measurable at all since it is just network communication and very low throughput.

S7.Net/PlcAsynchronous.cs
mycroes: The byte identifies PDU sizes (instead of giving a PDU value), this is from the RFC:
```
0000 1101  8192 octets (not allowed in Class 0)
0000 1100  4096 octets (not allowed in Class 0)
0000 1011  2048 octets
0000 1010  1024 octets
0000 1001   512 octets
0000 1000   256 octets
0000 0111   128 octets
```

Now take the decimal values of above, and perform 2 ^ [decimal value]:
```
binary     size          dec.   2 ^ dec.
0000 1101  8192 octets   13     8192
0000 1100  4096 octets   12     4096
0000 1011  2048 octets   11     2048
0000 1010  1024 octets   10     1024
0000 1001   512 octets    9      512
0000 1000   256 octets    8      256
0000 0111   128 octets    7      128
```

I actually looked into Sally7 to verify this and I must admit I was using the same logic as applied here (UInt 16 from 2 bytes), so honestly I have no clue whether the S7 protocol adheres to the RFC. I guess this is something to investigate.

The Connection Request is actually different from Data Transfer (which I believe is used for read/write), Connection Request mostly or entirely adheres to the RFC while read/write uses a custom protocol inside DT messages.

scamille: I changed it to just use uint16 big endian parsing for now. This also matches your current Sally7 implementation.

We should probably split further investigating of this into a separate issue. Might require some investigating with different PLC types or analyzing other programs using the S7 protocol.

S7.Net/PlcAsynchronous.cs
scamille: Well unless we figure out how exactly the PDU should be parsed, I am not adding a check. But we should remember this and once we are absolutely sure we can also check this lower limit.

S7.Net/PlcAsynchronous.cs
scamille: I opened #327 for this.

scamille:
I created #327 so we can further look into the MaxPDUSize parsing independent of this.

Is there anything else in here which still needs addressing?

mycroes (approved):

PR #316, opened 2020-09-12
Add more extensive response code validation.

### LINKED ISSUE #310: Improve Feedback when reading unavailable DB or reading too many bytes
When reading a DataBlock from the PLC that does not exist, or more bytes than exist in that DB, the error message is quite cryptic.

The idea is to figure out where/how in the S7 protocol such failures are communicated and turn that into a better Exception message.

scamille:
Fixes #310

### DIFF
--- a/S7.Net/PLC.cs
+++ b/S7.Net/PLC.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Sockets;
+using S7.Net.Protocol;
 using S7.Net.Types;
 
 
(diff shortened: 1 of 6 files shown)


S7.Net/PLC.cs
```diff
@@ -234,13 +235,34 @@ private static void AssertReadResponse(byte[] s7Data, int dataLength)
...
+                case ReadWriteErrorCode.DataTypeInconsistent:
+                    throw new Exception($"Received error from PLC: Data type inconsistent.");
+                case ReadWriteErrorCode.DataTypeNotSupported:
+                    throw new Exception($"Received error from PLC: Data type not supported.");
+                case ReadWriteErrorCode.AccessingObjectNotAllowed:
+                    throw new Exception($"Received error from PLC: Accessing object not allowed.");
+                case ReadWriteErrorCode.AddressOutOfRange:
+                    throw new Exception($"Received error from PLC: Address out of range.");
+                case ReadWriteErrorCode.HardwareFault:
+                    throw new Exception($"Received error from PLC: Hardware fault.");
```
mycroes, line 258: These messages don't need to be interpolated because they don't have any interpolations 😉

mycroes (approved):
Would personally remove the interpolations, other than that it's fine.

S7.Net/PLC.cs
scamille: Resolved in b95b71e

PR #317, opened 2020-09-12
Unify async read request header building.

scamille:
Add concept of DataRequestItem which contains the necessary data to build the request.

### DIFF
--- a/S7.Net/DataRequestItem.cs
+++ b/S7.Net/DataRequestItem.cs
@@ -1,0 +1,42 @@
+﻿
+using S7.Net.Types;
+
+namespace S7.Net
+{
+    internal class DataRequestItem
+    {
+        public DataRequestItem(DataType dataType, int db, int startByteAddress, int byteLength)
+        {
+            DataType = dataType;
+            DB = db;
+            StartByteAddress = startByteAddress;
+            ByteLength = byteLength;
+        }
+
+        public DataRequestItem(DataItem dataItem)
+            : this(dataItem.DataType, dataItem.DB, dataItem.StartByteAdr, Plc.VarTypeToByteLength(dataItem.VarType, dataItem.Count))
+        {
+
+        }
+
+        /// <summary>
+        /// Memory area to read 
+        /// </summary>
+        public DataType DataType { get; }
+
+        /// <summary>
+        /// Address of memory area to read (example: for DB1 this value is 1, for T45 this value is 45)
+        /// </summary>
+        public int DB { get; }
+
+        /// <summary>
+        /// Address of the first byte to read
+        /// </summary>
+        public int StartByteAddress { get; }
+
+        /// <summary>
+        /// Length of data to read
+        /// </summary>
+        public int ByteLength { get; }
+    }
+}
--- a/S7.Net/PlcAsynchronous.cs
+++ b/S7.Net/PlcAsynchronous.cs
@@ -7,6 +7,7 @@ using System.Threading.Tasks;
 using S7.Net.Protocol;
 using System.IO;
 using System.Threading;
+using S7.Net.Protocol.S7;
 
 namespace S7.Net
 {
@@ -262,7 +263,7 @@ namespace S7.Net
 
             try
             {
-                var dataToSend = BuildReadRequestPackage(dataItems.Select(d=> new DataRequestItem(d)).ToList());
+                var dataToSend = BuildReadRequestPackage(dataItems.Select(d => DataItem.GetDataItemAddress(d)).ToList());
                 await stream.WriteAsync(dataToSend, 0, dataToSend.Length);
 
                 var s7data = await COTP.TSDU.ReadAsync(stream, cancellationToken);
@@ -435,7 +436,7 @@ namespace S7.Net
         {
             var stream = GetStreamIfAvailable();
 
-            var dataToSend = BuildReadRequestPackage(new [] { new DataRequestItem(dataType, db, startByteAdr, count)});
+            var dataToSend = BuildReadRequestPackage(new [] { new DataItemAddress(dataType, db, startByteAdr, count)});
             await stream.WriteAsync(dataToSend, 0, dataToSend.Length, cancellationToken);
 
             var s7data = await COTP.TSDU.ReadAsync(stream, cancellationToken);
--- a/S7.Net/PlcAsynchronous.cs
+++ b/S7.Net/PlcAsynchronous.cs
@@ -285,19 +285,6 @@ namespace S7.Net
             return dataItems;
         }
 
-        private static byte[] BuildReadRequestPackage(IList<DataRequestItem> dataItems)
-        {                
-            // first create the header
-            int packageSize = 19 + (dataItems.Count * 12);
-            var package = new System.IO.MemoryStream(packageSize);
-            BuildHeaderPackage(package, dataItems.Count);
-            // package.Add(0x02);  // datenart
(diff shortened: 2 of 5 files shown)


S7.Net/PlcAsynchronous.cs
```diff
@@ -279,6 +269,20 @@ public async Task<List<DataItem>> ReadMultipleVarsAsync(List<DataItem> dataItems
...
+            int packageSize = 19 + (dataItems.Count * 12);
+            var package = new System.IO.MemoryStream(packageSize);
+            BuildHeaderPackage(package, dataItems.Count);
+            // package.Add(0x02);  // datenart
+            foreach (var dataItem in dataItems)
+            {
+                BuildReadDataRequestPackage(package, dataItem.DataType, dataItem.DB, dataItem.StartByteAddress, dataItem.ByteLength);
+            }
+            return package.ToArray();
+        }
```
mycroes, line 284: This _should_ be in PlcHelpers.cs, although in all honesty PlcHelpers.cs shouldn't exist at all. Also, one could say it'd be good to have methods like this available for testing, in which case they'd be better off in an internal class.

S7.Net/PlcAsynchronous.cs
```diff
@@ -279,6 +269,20 @@ public async Task<List<DataItem>> ReadMultipleVarsAsync(List<DataItem> dataItems
             return dataItems;
         }
 
+        private static byte[] BuildReadRequestPackage(IList<DataRequestItem> dataItems)
+        {                
+            // first create the header
+            int packageSize = 19 + (dataItems.Count * 12);
+            var package = new System.IO.MemoryStream(packageSize);
+            BuildHeaderPackage(package, dataItems.Count);
+            // package.Add(0x02);  // datenart
```
mycroes, line 278: Not sure why this comment is here, also I really dislike German (or any non-English) comments in code.

S7.Net/DataRequestItem.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// Address of the first byte to read
+        /// </summary>
+        public int StartByteAddress { get; }
+
+        /// <summary>
+        /// Length of data to read
+        /// </summary>
+        public int ByteLength { get; }
+    }
```
mycroes, line 41: I'm really not a fan of this. Even though it's not publicly exposed, it adds yet another slightly different version of something that's already there. The only difference being the `ByteLength` vs the `Count`, I'd say that if all use of `DataItem` for read requests is consolidated that we don't really need this class to (Dr. Evil quotes) 'fix' a shortcoming on `DataItem`.

I doubt if I would even mind having `ByteLength` exposed on `DataItem` (one could argue the importance of knowing this `ByteLength` because it's not so much a technical _detail_ as it would be for whatever item in whatever language where addressing is not important at all). Another option is to create an extension method for the ByteLength, which could either be internal or public.

mycroes (changes requested):
My comments are opinions, although I hope you share my point of view, I don't mind criticism if you don't agree!

S7.Net/DataRequestItem.cs
scamille: I really did not like reusing data item for this. It does have a lot of field not relevant for building the request header, and interprets some of them quite differently as well.

I do understand that this adds yet another class, but I think that is perfectly ok because it does represent exactly the data required to create a S7 read request.

I think we should mainly discuss how we can better organize things between the different layers and classes so that something like this clearly is a internal data structure for the S7 protocol.

Oh and I definitely did not want to mess with DataItem itself.

S7.Net/PlcAsynchronous.cs
scamille: Yeah I could have cleaned that up when copy-pasting. will do.

S7.Net/PlcAsynchronous.cs
scamille: Yeah you mentioned before that it would be great to separate concerns in the PLC functions.

I already made this static since it really does not require any PLC state.

I am also open for generally adding some folders/namespaces for all internal S7 protocol related stuff.

(Current Protocol folder has things which I don't think really belong in there)

S7.Net/DataRequestItem.cs
scamille: With Net5 this could be a record class :P
I think I will remove the constructor taking a dataItem to more clearly separate the two.

At least for me the DataItem right now is really just too much more than a mere S7 protocol data structure. It does serialization, both read and write, is mutable. Really not desirable to have for the internal stuff.

S7.Net/PlcAsynchronous.cs
mycroes: Wait, what?! You don't like my Protocol folder!? 😉 Yeah, even that's not coherent... Although some stuff is copied / refactored form older code, that's stuff I put there. WriteMultiple actually was one of my contributions before I became maintainer... But yeah, sure, let's start separating concerns into namespaces. Feel free to apply whatever you want, but you can also email me (or comment on the relevant PR, but you might not have a PR yet) if you want to discuss something first.

S7.Net/PlcAsynchronous.cs
mycroes: Didn't notice it was copy-paste, obviously you're not to blame for what someone else wrote.

S7.Net/DataRequestItem.cs
mycroes: You're absolutely right. I'm not a user (nor day to day developer) of S7NetPlus, so I didn't know about everything in `DataItem` either. I must say that records could bring welcome improvements to this codebase, also I don't think they'll be limited to Net5 either, AFAIK they're a C# 9 language feature and I don't see any reason why it wouldn't work on older targets.

All in all I'm not a fan of the name `DataRequestItem`. I agree with the structure and your vision, but maybe we can think of a better name for this? Basically this is just the address of a variable (or IO I guess, so perhaps variable is a bad term), not sure if `DataItemAddress` would rock my socks though.

scamille: If there was some documentation of the S7 protocol we could probably name it just like that...

You are right that Request should not necessarily be in there, since that is more the function of the message sent to the PLC, and does not really define the data here.

DataItemAddress does sound quite good and in lack of a better alternative from a S7 protocol description, I will change it to that.
-------------
Assuming records do not require anything special from the runtime, they could definitely be supported for older versions. C#8 is also not enabled by default for .NET Framework, but it is still very nice to use nullable annotations.

S7.Net/PlcAsynchronous.cs
scamille: I think a S7 folder inside the Protocol folder will do nicely. I have seen that in some other S7 library ;)

S7.Net/DataRequestItem.cs
mycroes: Are you still intending to rename this?

mycroes (commented):
I'm a bit short on time again, but I do intend to keep spending time at least a couple evenings a week on reviewing and going through issues. Let me know about the DataRequestItem / DataItemAddress name, I'm willing to merge this when that's settled.

scamille:
I am a bit short with time as well. Yes I do plan to rename this. I think I have already done something locally, but it needs some more work.

mycroes (approved):

mycroes:
Thanks again! I'm a bit too busy atm, so I'm short on time to handle all the PR's and issues....

PR #319, opened 2020-09-13
GitHub actions

scamille:
Depends on #318

Add github actions for multi-platform unit tests.


scamille:
It looks like dotnet restore just takes a lot longer on windows hosts than on the other ones. There is a ton of confusing information out there, nothing I found which sounds very applicable to our use case.
But since the overall run times are still quite small I just hope it will be fixed/improved eventually :)

mycroes:
I'm sorry for the delay on this. I will review and add this at some point, but I want to read a bit more about configuration before I do.

mycroes:
@scamille I just ran into the very reason that I'm hesitant to merge this: https://github.com/AutoMapper/AutoMapper.Extensions.ExpressionMapping/pull/91/checks?check_run_id=1255253798. That's the build from a PR I did, which probably fails because the GitHub action is badly configured (at least, that's my assumption). Nevertheless, it's also the first time I've actually seen this in action and I did like it, so once I have some time I'll try to enable this.

scamille:
I don't really see the problems there applying here that much.

I explicitly removed the artifact build step from the github actions step and the intention is to only use it for having a multi platform test setup.

Github actions are quite reliable, we use them at another project (https://github.com/simulationcraft/simc/actions) with both a larger workflow and much more commits.
That said, they sometimes do need a bit more maintenance and have bleeding edge features, but once set up correctly it usually works just fine.

Anyway: I still think it is worth merging this if only for testing S7NetPlus both with netcore and with other platforms. Keep Travis for the budget package and it shouldn't have any negative impact on the project.

scamille:
Please review again. Contains only the test workflow now.

scamille:
@mycroes Please review this again. I still think this is a sensible addition for S7NetPlus to increase test coverage to other platforms.

I removed any deployment workflow from the github actions, so this only contains test CI. The idea is to have this active alongside appveyor for now.

scamille:
Ok now finished with Net5.0 support, see https://github.com/scamille/s7netplus/actions/runs/909696878

scamille:
Somehow I both love and hate DevOps work at the same time. But there is good money to be made there, so maybe it can't hurt to get even more experience with it :)

mycroes (approved):

PR #335, opened 2020-09-28
Update PLC.cs

FalcoGoodbody:
update IsConnected() by sending a poll and check the received data

### DIFF
--- a/S7.Net/PLC.cs
+++ b/S7.Net/PLC.cs
@@ -107,11 +107,35 @@ namespace S7.Net
             {
                 try
                 {
+                    bool clientConnected = false;
+
                     if (tcpClient == null)
                         return false;
 
                     //TODO: Actually check communication by sending an empty TPDU
-                    return tcpClient.Connected;
+                    if (tcpClient.Client.Poll(0,SelectMode.SelectRead))
+                    {
+                        byte[] buffer = new byte[1];                        
+                        if (tcpClient.Client.Receive(buffer, SocketFlags.Peek) == 0)
+                        {
+                            // client disconnected
+                            clientConnected = false;                            
+                        }
+                        else
+                        {
+                            // client connected
+                            clientConnected = true;                            
+                        }
+                    }
+
+                    if (clientConnected)
+                    {
+                        return tcpClient.Connected;
+                    }
+                    else
+                    {
+                        return false;
+                    }
                 }
                 catch { return false; }
             }


scamille:
It is great to have a solution to actively check the connected status!

That said, I would not integrate it into the existing IsConnected function. It does pessimise the normal flow of execution with which works under the assumption that a simple Socket IsConnected check is sufficient - and the case where the connection is lost should be a exceptionally rare case.

Your solution would cause another send/receive round trip, which depending on the connection adds easily another 10-100ms.

S7.Net/PLC.cs
```diff
@@ -107,11 +107,35 @@ public bool IsConnected
...
 
                     //TODO: Actually check communication by sending an empty TPDU
-                    return tcpClient.Connected;
+                    if (tcpClient.Client.Poll(0,SelectMode.SelectRead))
+                    {
+                        byte[] buffer = new byte[1];                        
+                        if (tcpClient.Client.Receive(buffer, SocketFlags.Peek) == 0)
+                        {
+                            // client disconnected
+                            clientConnected = false;
```
scamille, line 122: you can just return false here, and skip the else branch. That also removes the need for the clientConnected variable.

S7.Net/PLC.cs
```diff
@@ -107,11 +107,35 @@ public bool IsConnected
...
                 try
                 {
+                    bool clientConnected = false;
+
                     if (tcpClient == null)
                         return false;
 
                     //TODO: Actually check communication by sending an empty TPDU
-                    return tcpClient.Connected;
+                    if (tcpClient.Client.Poll(0,SelectMode.SelectRead))
```
mycroes, line 116: Not sure if this tells us anything useful, true is also returned on closed sockets.

S7.Net/PLC.cs
```diff
@@ -107,11 +107,35 @@ public bool IsConnected
...
+                    {
+                        byte[] buffer = new byte[1];                        
+                        if (tcpClient.Client.Receive(buffer, SocketFlags.Peek) == 0)
```
mycroes, line 119: The receive method will block as long as it needs to read data, it only immediately returns when the socket has been closed. My (personal and professional) opinion: don't use the sync `Receive`, but use `ReceiveAsync` (if you need to use either that is). From this it follows that you can't use it in a property, which is fine, because you can't assert that communication will be successful anyway. Also, I thought I actually copied the comment from `TcpClient.IsConnected` to this `IsConnected` property (but apparently I didn't?), which clearly states that you can use it to see if the last communication attempt has been succesful and that you can't really rely on it anyway. I'd like to keep the IsConnected property as simple as that.

All this doesn't mean there's no room for improvement, but perhaps it would just be a good idea to fire a read for DB1,1 and see if a valid TPKT is returned (which could have an error indication in the message, but will still tell if the connection is fine).

mycroes (changes requested):
Not how I want this changed. An actual check should be async, as commented we could perhaps just do a read for whatever variable.

mycroes:
Actually, #311 has the IsConnected comment.

scamille:
In effect, what we would want is some function
public Task<bool> CheckIfConnected(CancellationToken);

Or even more verbose CheckIfConnectionisStillAlive or CheckifConnectionCanStillBeReadFrom. Whatever doesn't hurt too much when reading it while still hinting at what it does :D

All while keeping the existing IsConnected property like it is.

mycroes:
Exactly.

PR #337, opened 2020-10-01
Update License.txt

bipin4sing:


mycroes:
Not sure if this has any intentions, but I was assuming perhaps you would add some clarification to this PR. Assuming this was an error.

PR #342, opened 2020-10-21
Methods for IP-address handling

SmackyPappelroy:
Added a helper method for checking if a string is a valid IP-address string.
Added synchronous method for Pinging an IP-address.

### DIFF
--- a/S7.Net/PLCHelpers.cs
+++ b/S7.Net/PLCHelpers.cs
@@ -258,5 +258,24 @@ namespace S7.Net
 
             return package.ToArray();
         }
+        
+        /// <summary>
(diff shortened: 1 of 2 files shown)


mycroes:
I'm sorry @SmackyPappelroy, but this has no place in this library. Pinging is not related to the S7 protocol. Being able to ping (or not being able to ping) also doesn't guarantee much either. Also, there's no reason why an address would be an ip-address, any valid host (be it a FQDN or just a hostname) would be fine. Also, there are already methods in the BCL for parsing ip-addresses, I don't want to replicate that behavior in here either.

Nevertheless I want to thank you for taking the time and effort to contribute.

Issue #336, opened 2020-09-30
Close connection on Stream EOF or other exception


scamille:
@mycroes There have been some issues recently (#334 and #328) where the user had trouble to continue operating after communication problems occurs.

The solution from @FalcoGoodbody was to manually force Close() after certain exception, which seems to work fine for him.

The reason I am opening this issue is that I don't think we should burden the user with having to manually closing the PLC connection on exception cases. All the user should be required to do is have a ```if(!plc.IsConnected) plc.Connect();``` line before doing stuff, in my opinion.

My current feeling is that we should definitely close the exception on any stream EOF. In that case we can probably not really recover from the faulty state with the next request.

Maybe we should even close the connection on any IOException? But I am not really sure if that is overzealous and has too many false positives, where the connection would still have been ok to work with.

FalcoGoodbody:
Hello together,

I just want show a whole method I use to read values from one plc.
There is everything included which made the application real stable.

```csharp
private void ReadPlcKonti1()
{
    string EintragLog;
    try
    {
        if (!plcKonti1.IsConnected)
        {
            plcKonti1.Close();
            plcKonti1.Open();
        }

        if (plcKonti1.IsConnected)
        {
            K1Zaehlwert1 = ((uint)plcKonti1.Read("DB202.DBD0")).ConvertToInt();
            K1Zaehlwert2 = ((uint)plcKonti1.Read("DB202.DBD4")).ConvertToInt();
            K1Geschwindigkeit = ((uint)plcKonti1.Read("DB202.DBD8")).ConvertToFloat();
            K1aktuelleLaufmeter = ((uint)plcKonti1.Read("DB202.DBD14")).ConvertToFloat();

            #region Rollen
            var statusProduzierteRollen = plcKonti1.ReadBytes(DataType.DataBlock, 203, 0, 1);
            NeueRolleKonti1 = statusProduzierteRollen[0].SelectBit(0);

            if (NeueRolleKonti1)
            {
                plcKonti1.ReadClass(prodRolleKonti1, 203, 2);
                var tstring = plcKonti1.Read(DataType.DataBlock, 204, 0, VarType.StringEx, 50, 0);
                AuftragsnummerKonti1 = tstring.ToString();
                plcKonti1.WriteBit(DataType.DataBlock, 203, 0, 0, false);
            }
            #endregion
        }                
    }
    catch (PlcException e)
    {                
        // wenn dies nicht hilft, dann Log-Eintrag und weiter
[...]

PR #349, opened 2020-11-11
Close connection on IOExceptions.

scamille:
Let our protocol exceptions derive from IOException so their errors are handled nicely alongside other IOExceptions.

The idea behind this is that if we receive a response which does not conform to the underlying protocol specification (with all the TCP checks in place!), we are in a weird/corrupt state and it is better to just close the connection.

### DIFF
--- a/S7.Net/PLC.cs
+++ b/S7.Net/PLC.cs
@@ -162,6 +162,7 @@ namespace S7.Net
             if (tcpClient != null)
             {
                 if (tcpClient.Connected) tcpClient.Close();
+                tcpClient = null; // Can not reuse TcpClient once connection gets closed.
             }
         }
 
--- a/S7.Net/PlcAsynchronous.cs
+++ b/S7.Net/PlcAsynchronous.cs
@@ -37,7 +37,7 @@ namespace S7.Net
                     return default(object);
                 }).ConfigureAwait(false);
             }
-            catch(Exception)
+            catch (Exception)
             {
                 stream.Dispose();
                 throw;
@@ -438,7 +438,7 @@ namespace S7.Net
 
         private async Task ReadBytesWithSingleRequestAsync(DataType dataType, int db, int startByteAdr, byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            var dataToSend = BuildReadRequestPackage(new [] { new DataItemAddress(dataType, db, startByteAdr, count)});
+            var dataToSend = BuildReadRequestPackage(new[] { new DataItemAddress(dataType, db, startByteAdr, count) });
 
             var s7data = await RequestTsduAsync(dataToSend, cancellationToken);
             AssertReadResponse(s7data, count);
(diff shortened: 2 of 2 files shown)


scamille:
This is about #336

S7.Net/PLCExceptions.cs
```diff
@@ -69,7 +70,7 @@ protected InvalidVariableTypeException(SerializationInfo info, StreamingContext
         #endif
     }
 
-    internal class TPKTInvalidException : Exception
+    internal class TPKTInvalidException : IOException
```
mycroes, line 73: I don't agree on this. An `IOException` in my opionion is an exception because in or out failed, in this case it didn't fail, it just returned something unexpected. It also doesn't matter much what the underlying exception was, when we can't read the TPKT or TPDU the connection is messed up and we need to close it. It would also help if the connection send/receive was used in a single place so we could standardize there.

mycroes (changes requested):
I don't agree to this change, even though the intention is good and it does solve a real problem. I'm not sure of the path forward, personally I'd rather have a single point of interaction with the `NetworkStream` before things like this are added.

S7.Net/PLCExceptions.cs
scamille: I agree with your argumentation, and will revert this to inherit from System.Exception.

In effect this means that I also need to check for those 2 exception in the places where I close the connection - since as you said the S7 connection is somehow messed up. As you said it would be great if there were fewer places where this would have to happen, but I don't think one can hope for the code to clean itself up all by itself, so in the meantime there would be a bit of duplication.

I guess the most likely use case for all of this is when people use a plc connection in a unsynchronized way from multiple threads. So maybe all of this is not worth it at all :)

scamille:
Now with a different logic, closing on invalid tpdu / tpkt exceptions.

What do you think? Does that make it worse, or is it what we are looking for?

mycroes (approved):

scamille:
Rebased this. Let me know if this is still something we want to include in its current form, or if it should be adjusted somehow after your changes.

S7.Net/PLCExceptions.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.IO;
```
mycroes, line 2: Leftover addition?

mycroes (commented):
Started a review, but basically, yes, this needs refactoring. Since (almost) all TPKT reading now is done in one place, that's the place to close the connection as well.

If you feel like nailing it, you could add `CancellationToken` support as well, by means of something along the lines of the following code:

```csharp
// ...

// This one is for cancelling the request when cancel actually happened before the request was dequeued
cancellationToken.ThrowIfCancellationRequested();

// When cancelling here we need to keep in mind data has been sent, so the connection is borked anyway.
using (cancellationToken.Register(() => conn.Close(), false) {
    await stream WriteAsync(/* ... */);
    await stream.ReadAsync(/* ... */);
}
```

Regards,
Michael

scamille:
Ok I'll adjust. Should make it much less invasive to only put the check into the new Tpdu and Tsdu functions.

mycroes (approved):

mycroes:
test-macos-latest-net5.0 sometimes has `OperationCanceledException` instead of `TaskCanceledException`, ignoring that for now...

scamille:
Reading https://stackoverflow.com/questions/34359129/operationcanceledexception-vs-taskcanceledexception-when-task-is-canceled explains why the types can differ. I think it might sense to just check for operation canceled exception, which is base for task cancelled exception anyway.

scamille:
Do you see a easy way to add a unit test for this?

The only somewhat feasible option without refactoring the PLC class that I see is using a fake stream that triggers a cancellation source cancel once some bytes have been written, which in turn can cancel the plc read/write operation while it is sending data. That should then result in my new code closing the stream.

mycroes:
I like your approach for the test, it's probably what I would do. Also, I don't mind opening up more internals of the library to ease unit testing, like publicly exposing the `TcpClient`. That might serve other uses as well. I think I also exposed it in Sally7.

PR #362, opened 2021-01-16
Net5

HakamFostok:
The PR contains
- Target .NET 5
- Updating NuGet Packages
- Formatting the code

I do NOT want to create this PR towards the "develop " branch, Actually, I wanted to create a new branch for .NET 5 and make the PR toward it.

if you can please create a branch like this and close this PR, I will reopen the PR to that new create .NET 5 branch.

I think this library is so great that could have a version specific for .NET 5

Yesterday, when I develop an application for my customer, which was forced to use .netcore2.2 application, I really needed to use this library, but it was not compatible with .netcore2.2, so I created a special build targeting .netcore2.2 and we solved a very big problem in the application by using this great library.

I have a lot of ideas about using C# 9 features in the library, I think this will increase the readability of the library.

By the way, I wanted to thank all of you for these great efforts which produced this great library, I wanted to pay back the debt I felt towards this library.

### DIFF
--- a/S7.Net.UnitTest/Helpers/ConsoleManager.cs
+++ b/S7.Net.UnitTest/Helpers/ConsoleManager.cs
@@ -70,10 +70,10 @@ namespace S7.Net.UnitTest.Helpers
             System.Reflection.MethodInfo _InitializeStdOutError = type.GetMethod("InitializeStdOutError",
                 System.Reflection.BindingFlags.Static | System.Reflection.BindingFlags.NonPublic);
 
-            System.Diagnostics.Debug.Assert(_out is not null);
-            System.Diagnostics.Debug.Assert(_error is not null);
+            System.Diagnostics.Debug.Assert(_out != null);
+            System.Diagnostics.Debug.Assert(_error != null);
 
-            System.Diagnostics.Debug.Assert(_InitializeStdOutError is not null);
+            System.Diagnostics.Debug.Assert(_InitializeStdOutError != null);
 
             _out.SetValue(null, null);
             _error.SetValue(null, null);
--- a/S7.Net.UnitTest/Helpers/ConsoleManager.cs
+++ b/S7.Net.UnitTest/Helpers/ConsoleManager.cs
@@ -70,10 +70,10 @@ namespace S7.Net.UnitTest.Helpers
             System.Reflection.MethodInfo _InitializeStdOutError = type.GetMethod("InitializeStdOutError",
                 System.Reflection.BindingFlags.Static | System.Reflection.BindingFlags.NonPublic);
 
-            System.Diagnostics.Debug.Assert(_out != null);
-            System.Diagnostics.Debug.Assert(_error != null);
+            System.Diagnostics.Debug.Assert(_out is not null);
+            System.Diagnostics.Debug.Assert(_error is not null);
 
-            System.Diagnostics.Debug.Assert(_InitializeStdOutError != null);
+            System.Diagnostics.Debug.Assert(_InitializeStdOutError is not null);
 
             _out.SetValue(null, null);
             _error.SetValue(null, null);
--- a/S7.Net.UnitTest/Helpers/S7TestServer.cs
+++ b/S7.Net.UnitTest/Helpers/S7TestServer.cs
@@ -87,7 +87,7 @@ namespace S7.Net.UnitTest.Helpers
             //        low and so it's considered "privileged".
         }
 
-        public static void Stop() 
+        public static void Stop()
         {
             int Error = Server.Stop();
         }
--- a/S7.Net.UnitTest/Helpers/TestLongClass.cs
+++ b/S7.Net.UnitTest/Helpers/TestLongClass.cs
@@ -4,7 +4,7 @@ using System.Linq;
 using System.Text;
 
 namespace S7.UnitTest.Helpers
-{ 
+{
     /// <summary>
     /// This is a class that contains more than 200 bytes and that needs 2 plc requests to complete a read/write cycle
     /// </summary>
--- a/S7.Net.UnitTest/S7.Net.UnitTest.csproj
+++ b/S7.Net.UnitTest/S7.Net.UnitTest.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net452;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net452;netcoreapp3.1;net5.0</TargetFrameworks>
 
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>Properties\S7.Net.snk</AssemblyOriginatorKeyFile>
--- a/S7.Net.UnitTest/S7.Net.UnitTest.csproj
+++ b/S7.Net.UnitTest/S7.Net.UnitTest.csproj
@@ -11,11 +11,11 @@
   </PropertyGroup>
   
   <ItemGroup>
-    <PackageReference Include="GitHubActionsTestLogger" Version="1.1.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.0" />
+    <PackageReference Include="GitHubActionsTestLogger" Version="1.1.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.8.3" />
     <PackageReference Include="MSTest.TestAdapter" Version="2.1.2" />
     <PackageReference Include="MSTest.TestFramework" Version="2.1.2" />
-    <PackageReference Include="coverlet.collector" Version="1.3.0">
+    <PackageReference Include="coverlet.collector" Version="3.0.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
--- a/S7.Net.UnitTest/S7.Net.UnitTest.csproj
+++ b/S7.Net.UnitTest/S7.Net.UnitTest.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net452;netcoreapp3.1</TargetFrameworks>
+	<TargetFramework>net5.0</TargetFramework>
 
     <SignAssembly>true</SignAssembly>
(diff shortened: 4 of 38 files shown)


scamille:
Is there a reason to not just add net5 as another target framework, instead of replacing the existing ones?
The even better choice might be to eventually just switch to netstandard2.0

I am not really the project owner here - but I can imagine that there is not that much interest in having a separate fork/branch for net5 which then needs to be maintained separately. Instead, having multiple target frameworks allows for all of them to be developed together in a single codebase. And even though netstandard2.1 would offers some newer network APIs as well, I don't think there is that much to gain/demand for them here, and compatibility with net framework and ease of maintenance is more important anyway.

HakamFostok:
After a second thought, I think you are right,
> newer network APIs as well, I don't think there is that much to gain/demand

I agree,

the only (may or may NOT) be a problem is that C# 9 features are only supported on .NET 5, even .NET standard 2.1 does not support C# 9,
we need to stick with C# 8

image

Maybe for time being, it is ok, but I think in the future versions S7NetPlus should support C# 9

So you think we can just add .NET 5 to the already supported languages?
Ok, my goal was just to expand the support of this library to the users who are targeting .NET Core, and I think this will achieve that?

I think we are on the same page, Right?

scamille:
My personal opinion would be to just add 'net5.0' to the <TargetFrameworks> element like you now did. It would probably also make a lot of sense to adjust and finish #319 so that both net core and net5.0 can get some automated testing for both net core 3.1 and net5.

It might also make sense to split up this PR into only adding net5 support and applying all those style & formatting changes in a separate PR. But all that is really up to @mycroes . Your opinion and decision on this is definitely required at some point :)

About C#9: It might be worth checking out which parts of C#9 are still supported by older frameworks, and which absolutely require NET 5. C#8 isn't supported by netstandard2.0 as well, but parts of it can still be without problems, for example nullable reference annotations.

HakamFostok:
> C#8 isn't supported by netstandard2.0

this brings the following question, why not also add .NET standard 2.1 ? is there any technical challenge for doing that?

mycroes:
Time to chime in for me I guess 😉 Not sure where you guys are getting the support frameworks for C# language versions, but we can perfectly use C# 9 features without targeting .NET 5. Also, we're not getting anything from targeting .NET 5 at this point, because we don't use any features that are not available on older specifications. I believe .NET 5 also supports all netstandard versions, so I'm not sure why you thought you needed to add .NET 5 support in the first place.

Also, Microsoft has once recommended you target the minimal spec that supports all features of your library. In our case it was easy to support both netstandard1.3 and 'something newer' (there's some conditional compilation, but not much), which means people can actually use this library on platforms that support up to netstandard1.3. I'm guessing those platforms might already be without upstream support nowadays anyways, so I'm not necessarily aiming to keep supporting netstandard1.3 for years to come. However, I definitely want to keep supporting .NET Framework (preferably something along the lines of 4.5.2) and .NET Core 2 unless there's a very good reason to no longer offer that support.

As for netstandard2.1, I'm not sure what improved network API would be included there. I honestly won't have noticed, because I don't develop against the network API regularly. If there is such an improved API that would also means everything below netstandard2.1 would require a different code path to support the older API, and I'm not sure that improved API would warrant dropping support for everything below netstandard2.1.

Anyway, that's my 2 cents about the framework targeting, now on to the other changes. I wouldn't mind a C#-9-cleanup-PR, which uses C# 9 features to improve code readability. I don't like the format-some-project commits though, for a couple of reasons. First, I would pick some change and apply it to the codebase, instead of having multiple changes applied on a per-project basis. Second, I'd rather have casts spaced out of the operand. I know it's currently a mess in the codebase (both with and without spaces), but if that's touched then it should be including a space. Other than that, feel free to use any C# 9 feature that improves the code, but as I said you should split such changes on feature-level rather than project level (so 'Replace == null with is null' is fine with regards to scope of the commit).

Regards,
Michael

HakamFostok:
I am really so busy because I am moving to another city, and I just have time on weekends.

To be honest I was surprised with much (if not all) of what you said.

 1. C# 9 can NOT be used in this library, see the following line in the library`<LangVersion>8.0</LangVersion>` and see the image I posted above, it is from Microsoft,  [here is the source](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version)
 
 2. it is not just about the improvements of Networking API, I agree with you maybe the Networking API has no improved, and let us suppose that the improvements on Networking API are zero. 
 It is not about the Networking API, it is about supporting more users, what made be landing here in the first place.
The problem I faced with my customer is, he is forced to use .NET Core 2.2 (no other choices), it is a giant car manufacturer with more than 10,000 employees.
when I tested (from visual studio) this library it worked well, everything ok, when they published it on their environment (Dockerized Environment), it start to fail (Argument Exception, Maybe I should have open an issue of that, but it was a lot of pressure and I just wanted it to work), what I did to solve the problem was exactly forking the library and build it against .CORE 2.2 (by the way, when I did that the Nullable References started to give compile-time errors because .NET core 2.2 support only C# 7.3, so I had to remove them) and I used that custom-build library, and after than every-one continued a life full of happiness.

3. >I definitely want to keep supporting .NET Framework

Did I suggest implicitly removing the support for it? of course, .NET Framework and .NET CORE 2.2 should be supported.

4. >I believe .NET 5 also supports all netstandard versions

I am not sure if I understand you correctly here, but if we took your words for the letters, I think that means Targeting .NET 5 is enough to support netstandard.

HakamFostok:
anyway, I am closing this PR, because I think it will not be approved.
wish you all the best, it is a great library, It helped me a lot, and you all are great guys.
nice to meet you,
have a nice day.

mycroes:
> I am really so busy because I am moving to another city, and I just have time on weekends.

No worries, I'm time constrained as well.

> 1. C# 9 can NOT be used in this library, see the following line in the library`<LangVersion>8.0</LangVersion>` and see the image I posted above, it is from Microsoft,  [here is the source](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version)

Yes, `<LangVersion>8.0</LangVersion>` restricts feature usage to C# 8.0 features, but there's no reason why we wouldn't change it to `9.0` or `latest` even. I'm not sure you're interpreting the Microsoft docs correctly; it lists the *default* LangVersion for target frameworks, but LangVersion doesn't limit the target frameworks to those defaults (how else would anyone be able to target `netstandard1.6` while using `<LangVersion>8.0</LangVersion>`?

> 2. it is not just about the improvements of Networking API, I agree with you maybe the Networking API has no improved, and let us suppose that the improvements on Networking API are zero.
>    It is not about the Networking API, it is about supporting more users, what made be landing here in the first place.

Actually it was @scamille that was hinting at an improved network API, I only took that as an example to indicate how such usage (of features available in newer target frameworks only) would cause either more work on maintaining the library or would limit the environments we can target.

>    The problem I faced with my customer is, he is forced to use .NET Core 2.2 (no other choices), it is a giant car manufacturer with more than 10,000 employees.
>    when I tested (from visual studio) this library it worked well, everything ok, when they published it on their environment (Dockerized Environment), it start to fail (Argument Exception, Maybe I should have open an issue of that, but it was a lot of pressure and I just wanted it to work), what I did to solve the problem was exactly forking the library and build it against .CORE 2.2 (by the way, when I did that the Nullable References started to give compile-time errors because .NET core 2.2 support only C# 7.3, so I had to remove them) and I used that custom-build library, and after than every-one continued a life full of happiness.

1. The ArgumentException is probably related to the default timeout, which I once set to `Timeout.Infinite`, which at least didn't work on Mono, but possibly the same issues exists for .NET Core. I can't be sure, but the issue has been fixed for the Mono runtime and recently someone was developing a mobile application using Xamarin and there it has also been fixed.
2. .NET Core *runtime* (not *compiler*) doesn't support _any_ C# version, it's a runtime, it takes MSIL and executes that. It's the compiler that takes care of changing C# to relevant MSIL instructions. As such, it's the compiler that needs to support the C# language features. Of course that means that if you're using an older compiler that it will constrain the C# language features you can use. I'm not 100% sure about the exact implementation of nullable reference types, but I believe it consists of attributes only to hint to the compiler whether something can be null, which means that you can even use a library with nullable reference type attributes without the compiler having any notion of these attributes, it'll just mean it can't provide the hinting that newer compilers can do.
3. You're running your software in a Docker container, that doesn't technically limit you to .NET Core 2.2. I can understand company policies might constrain you (which is fine), but there's no reason you can't use S7NetPlus in that environment.
4. You're talking about _custom-build library_, did you originally use S7NetPlus as NuGet package or did you include the source in your application? Some things you're saying seem to hint at the latter combined with a development or deployment environment that doesn't support recent C# features.
5. Besides maintainer of S7NetPlus (which is not my library originally), I'm also the creator of [Sally7](/mycroes/Sally7), which has a simpler API, higher performance and a lot less code (to audit, digest, whatever). It does lack some features that S7NetPlus supports, because I've never needed those professionally.

> 3. > I definitely want to keep supporting .NET Framework
> 
> Did I suggest implicitly removing the support for it? of course, .NET Framework and .NET CORE 2.2 should be supported.

No, but you seem to lack understanding of multi-targeting and compatibility in general, so I made a minor attempt at explaining it.

> 
> 1. > I believe .NET 5 also supports all netstandard versions
> 
> I am not sure if I understand you correctly here, but if we took your words for the letters, I think that means Targeting .NET 5 is enough to support netstandard.

You got it the wrong way around. .NET Framework, .NET Core and .NET 5 are runtimes. [.NET Standard is a formal specification of .NET APIs](https://docs.microsoft.com/en-us/dotnet/standard/net-standard). Because we target netstandard1.3 as well you're able to use our library on any runtime supporting at least netstandard1.3. We also target some newer netstandard versions and conditionally use some features not available in netstandard1.3, so those features are available on all runtimes that support those netstandard versions. I'm not sure if this is enough of an explanation to understand how this works, but otherwise I can definitely recommend reading up on this (especially if this is your day to day work!).

> anyway, I am closing this PR, because I think it will not be approved.

That's stating the obvious, I said I wouldn't approve it in it's current form due to multiple highlighted concerns.

> nice to meet you,

You forgot the _sarcasm_ tags. Honestly, you're coming in like you're the big shot developer (working at a giant car company with 10.000 employees), but I really hope for the company they have experienced developers as well. You've got a shitty attitude that will earn you rejections like on this PR. Somehow I feel I need to point this out to you, I guess that makes me an asshole as well.

HakamFostok:
Thank you, Sir.
I learned a lot from that, and I will continue to read to improve my knowledge.
Really sorry for the disturbance that I caused.
I am not a big-shot or anything, I am trying to learn, I work in a very small company with 10 people, the car company is a customer for the company I am working in.
Sorry again for everything.
Have a nice day.

PR #365, opened 2021-01-29
Added support for string types in a struct

MCPC10:
- Added support for string types in a struct.

### DIFF
--- a/S7.Net/Types/S7StringAttribute.cs
+++ b/S7.Net/Types/S7StringAttribute.cs
@@ -57,11 +57,11 @@ namespace S7.Net.Types
         /// <summary>
         /// ASCII string.
         /// </summary>
-        S7String,
+        S7String = VarType.S7String,
 
         /// <summary>
         /// Unicode string.
         /// </summary>
-        S7WString
+        S7WString = VarType.S7WString
     }
 }
--- a/S7.Net/Types/S7StringAttribute.cs
+++ b/S7.Net/Types/S7StringAttribute.cs
@@ -1,0 +1,67 @@
+﻿using System;
+
+namespace S7.Net.Types
+{
+    [AttributeUsage(AttributeTargets.Field, AllowMultiple = false)]
+    public sealed class S7StringAttribute : Attribute
+    {
+        private readonly S7StringType type;
+        private readonly int reservedLength;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="S7StringAttribute"/> class.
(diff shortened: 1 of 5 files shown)


S7.Net/Types/Struct.cs
```diff
@@ -178,6 +189,32 @@ public static int GetStructSize(Type structType)
...
+                    case "String":
+                        S7StringAttribute? attribute = info.GetCustomAttributes<S7StringAttribute>().SingleOrDefault();
+                        if (attribute == default(S7StringAttribute))
+                            throw new ArgumentException("Please add S7StringAttribute to the string field");
+
+                        numBytes = Math.Ceiling(numBytes);
+                        if ((numBytes / 2 - Math.Floor(numBytes / 2.0)) > 0)
+                            numBytes++;
+
+                        // hier auswerten
```
mycroes, line 201: Would prefer English comments only 😀

S7.Net/Types/S7StringAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+        private readonly S7StringType type;
+        private readonly int reservedLength;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="S7StringAttribute"/> class.
+        /// </summary>
+        /// <param name="type">The string type.</param>
+        /// <param name="reservedLength">Reserved length of the string in characters (default value is 254).</param>
+        /// <exception cref="ArgumentException">Please use a valid value for the string type</exception>
+        public S7StringAttribute(S7StringType type, int reservedLength = 254)
```
mycroes, line 17: Not a fan of the default length. Is this also the default when you add a string in TIA Portal? Personally I would prefer that you'd have to be explicit about this, so that you can't forget either.

mycroes (commented):
I'm sorry for the late reply, but this is a very nice and clean change. I added two minor comments, I hope you can answer about the string length and then I'd be happy to merge this.

S7.Net/Types/Struct.cs
MCPC10: Ups, just copied/inserted the code above without changing the comment xD

S7.Net/Types/S7StringAttribute.cs
MCPC10: I made it like this because it is normally the default length of a string (also wstring) type in TIA. But you are right, that it is probably better to make it explicit.

mycroes (approved):

Issue #353, opened 2020-12-06
Autoreconnect


sonduter:
Hi everyone. I have a project with Visual C# and mutiples s7_1200. I have a problem when one of PLCs disconnects because of ưeak network then my application not working.

Issue #234, opened 2019-09-10
Any way to add DateTime and string support to ReadStruct / ReadClass?


Equino3108:
I realized strings and datetime are not supported in ReadStruct / ReadClass Methods.
Is there a way to get a datetime field into These structs/classes ? My first idea was to try and read them as strings, but since they aren't supported either?

Thanks

FalcoGoodbody:
Hi @Equino3108,

it is possible to read strings from a plc. Check the issue #239 . There is an example provided.

tolgacakir:
Hi,
`ReadStruct`/ `ReadClass` methods do not support **strings**, it should be written to the **document** or **wiki** page.
Will `ReadStruct`/ `ReadClass` methods support **strings**?

scamille:
#365 would also enable support for string in struct/classes. Note that this requires adding annotating your string type and capacity, since that can't be inferred from your string alone.

Datetime support in struct/class would also require some annotations to tell S7NetPlus which S7 datetime type to use (there are multiple). And anyway, someone would have to implement it.

You can of course use the same code as @FalcoGoodbody provided just with VarType.DateTime and then cast to (DateTime).

mycroes:
I'm leaving this issue open for now (#365 will address the string part when merged), there's no reason why DateTime couldn't be implemented in a similar fashion to how #365 implements string reading (except for the fact that it's simply not necessary to specify a length for DateTime).

PR #367, opened 2021-02-09
Added read/write class char array support and the auto reconnector support for PLC

tolgacakir:
Hi,
This pull request includes the array of char and string support for the ReadClass / WriteClass #234 
And includes an auto reconnector mechanism for the PLC  #353

### DIFF
--- a/S7.Net/AutoReconnector.cs
+++ b/S7.Net/AutoReconnector.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Timers;
+
+namespace S7.Net
+{
+    public class AutoReconnector
+    {
+        private readonly Plc _plc;
+        private readonly Timer _timer;
+        public AutoReconnector(Plc plc, double interval = 3000)
+        {
+            _plc = plc;
+            _timer = new Timer(interval);
+            _timer.Elapsed += _timer_Elapsed;
+        }
+        public void Start()
+        {
+            _timer.Start();
+        }
+
+        public void Stop()
+        {
+            _timer.Stop();
+        }
+
+        private void _timer_Elapsed(object sender, ElapsedEventArgs e)
+        {
+            if (!_plc.IsConnected)
+            {
+                _plc?.Open();
+            }
+        }
+        
+    }
+}
--- a/S7.Net.UnitTest/Helpers/TestClass.cs
+++ b/S7.Net.UnitTest/Helpers/TestClass.cs
@@ -51,5 +51,12 @@ namespace S7.Net.UnitTest.Helpers
         /// DB1.DBD16
         /// </summary>
         public ushort DWordVariable { get; set; }
+
+        /// <summary>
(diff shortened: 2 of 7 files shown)


S7.Net/AutoReconnector.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            _timer.Start();
+        }
+
+        public void Stop()
+        {
+            _timer.Stop();
+        }
+
+        private void _timer_Elapsed(object sender, ElapsedEventArgs e)
```
mycroes, line 30: This doesn't adhere to coding style used in this project.

S7.Net/AutoReconnector.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            if (!_plc.IsConnected)
+            {
+                _plc?.Open();
```
mycroes, line 34: Why the nullability check? If it was null, you'd probably already have an exception on the line above. Also, if it would ever be null here, that means it was initially passed as null into the constructor, so this doesn't make any sense to me.

S7.Net/AutoReconnector.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Timers;
```
mycroes, line 6: Not sure why you used `System.Timers.Timer`, but I think `System.Threading.Timer` is the better choice in this case.

S7.Net/AutoReconnector.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace S7.Net
+{
+    public class AutoReconnector
```
mycroes, line 10: The way this is setup is a bad idea. The PLC class is not threadsafe, this will perform `Open` calls from a different thread, it's broken fundamentally.

mycroes:
Hi @tolgacakir,

Sorry for this very late reply. To be honest, I was already put off by the title of the PR so that didn't help in getting me to review this. Now I guess you might be wondering 'what could possibly be wrong with the title?', so let's start there. This PR combines two things. I don't like combined changes or combined PR's. This should have been a PR for the 'AutoReconnector' and a PR for the char array support.

As for the AutoReconnector support, I've provided some comments to the code, but let me put it this way, I won't merge it in this form.

On to the char array support... A char in C# is UTF-8 encoded, as such the length in bytes is not fixed (it actually uses 1 to 4 bytes per character). Simply put, your conversion is incorrect. Next issue, you're just dumping the array into the data to write. There's no length configuration, in any way. Even though that could work if you take care to zero-pad the char array if less characters are present than the reserved size in the PLC, this is just a very bad idea... In other words, this won't get in in this form either. I'm not sure if there's even support for UTF-8 strings in Siemens PLC's.

Nevertheless I want to thank you for the time taken to create this PR.

S7.Net/AutoReconnector.cs
paviano: Not just a better choice, probably the only real one 😊

PR #377, opened 2021-03-15
Fix wrong calculation for S7WString byte length

### LINKED ISSUE #376: Bad ByteLength for S7Strings and S7WStrings

ismdiego:
Fixes #376


ismdiego:
I will reopen a new one because I have found another problem with S7String

PR #378, opened 2021-03-15
Fix problems with S7 string types (S7String and S7WString) and calculated byte length

### LINKED ISSUE #376: Bad ByteLength for S7Strings and S7WStrings

ismdiego:
This pull requests fixes #376.

The S7WString fix it is a simple one with no danger of breaking anything. The S7String fix can be more dangerous, although it passes all the tests.

I also added two new test methods to both test cases related to these fixes.

### DIFF
--- a/S7.Net.UnitTest/TypeTests/S7StringTests.cs
+++ b/S7.Net.UnitTest/TypeTests/S7StringTests.cs
@@ -117,12 +117,23 @@ namespace S7.Net.UnitTest.TypeTests
             AssertToByteArrayAndBackEquals("Abc", 4, 4, 3, (byte) 'A', (byte) 'b', (byte) 'c', 0);
         }
 
+        [TestMethod]
+        public void OddS7StringByteLength()
+        {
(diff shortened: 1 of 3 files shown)


mycroes:
Hi @ismdiego, thanks for your contribution. I'm not sure what's up with AppVeyor not building this (well, it's GitVersion that couldn't establish a version, but still I'm not sure why that's happening), but the code looks good to me. I'll try to fix CI somewhere soon so this can be merged.

ismdiego:
> Hi @ismdiego, thanks for your contribution. I'm not sure what's up with AppVeyor not building this (well, it's GitVersion that couldn't establish a version, but still I'm not sure why that's happening), but the code looks good to me. I'll try to fix CI somewhere soon so this can be merged.

Hi @mycroes, anytime. If you cannot fix CI with this pull request anytime soon, can you manually recode it in the main branch and cancel this one? I have no previous experience with pull requests and maybe I did something wrong.
I would like to use the library as a NuGet Package and this bug prevents me from doing so (as I need S7WStrings) =)
Thank you very much

PR #381, opened 2021-03-18
Check maximum number of data items that can be read in a single request.

### LINKED ISSUE #380: Using ReadMultipleVars with more than 20 data items throws an index out of range exception
When using more than 20 'DataItem' objects in a 'ReadMultipleVarsAsync' request, it throws the following exception:

`S7.Net.PlcException: Index was outside the bounds of the array. ---> System.IndexOutOfRangeException: Index was outside the bounds of the array. at S7.Net.Plc.ParseDataIntoDataItems(Byte[] s7data, List`1 dataItems)`

All 'DataItems' were of type 'VarType.Bit'.

Is this a known limitation? I couldn't find anyfhing about it in the wiki.

scamille:
Fixes #380

### DIFF
--- a/S7.Net.UnitTest/S7NetTestsAsync.cs
+++ b/S7.Net.UnitTest/S7NetTestsAsync.cs
@@ -1,6 +1,7 @@
 ﻿#region Using
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using S7.Net.UnitTest.Helpers;
 using S7.Net.Types;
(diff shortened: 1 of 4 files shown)


mycroes:
I'm not sure about this change. I originally added the request item constraints based on Siemens documentation, I think they're been altered later because someone thought they were incorrect. I'm gonna check the documentation again (at some point) and will get back to this later.

mycroes:
I'm quite sure 20 is not the limit, even though it is the limit for Snap7 as server I think. It seems like the original code is correct, so I'm closing this one.

PR #382, opened 2021-03-29
Fix problems with S7 string types (S7String and S7WString) and calculated byte length

### LINKED ISSUE #376: Bad ByteLength for S7Strings and S7WStrings

mycroes:
New upload of #378, in hopes of working around the CI issues there.

This pull requests fixes #376.

The S7WString fix it is a simple one with no danger of breaking anything. The S7String fix can be more dangerous, although it passes all the tests.

I also added two new test methods to both test cases related to these fixes.

### DIFF
--- a/S7.Net.UnitTest/TypeTests/S7StringTests.cs
+++ b/S7.Net.UnitTest/TypeTests/S7StringTests.cs
@@ -117,12 +117,23 @@ namespace S7.Net.UnitTest.TypeTests
             AssertToByteArrayAndBackEquals("Abc", 4, 4, 3, (byte) 'A', (byte) 'b', (byte) 'c', 0);
         }
 
+        [TestMethod]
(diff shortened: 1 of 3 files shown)


mycroes:
@ismdiego Thanks for the changes, merging as is!

ismdiego:
You are welcome, I think the problem with the CI was caused because I did not create a named branch or label or whatever is called :-) I seldom use git (I work with hg instead), but I have to learn to not make this kind of mistakes.

Thank you 👍

mycroes:
@ismdiego This definitely wasn't your fault. GitVersion is to blame. It serves a good purpose but it has been giving me a lot of pain in the process as well, both for professional and personal projects. Also, the branching strategy I chose might not be the best choice for S7NetPlus (or my other projects), that doesn't help for newcomers. Thanks again for your effort though!

PR #388, opened 2021-04-30
Master

Jason-Jelks:
Per Siemens documentation ( https://support.industry.siemens.com/cs/document/43566349/in-step-7-(tia-portal)-how-can-you-input-read-out-and-edit-the-date-and-time-for-the-cpu-modules-?dti=0&lc=en-WW )
DateTime & DateTimeLong are not interchangeable between all Siemens PlcTypes.
DT => S7200, S7300 & S7400
DTL -> S71200 & S71500

Also, the S7.Class object does not support a String data type in the .net class object.


Jason-Jelks:
mycroes, I hope I posted this to the correct fork. For my testing, I am using actual Siemens S7-400, S7-1200, and S7-1500 PLCs. The DateTime / DateTimeLong changes made were technically correct, but the parsing causes problems when you attempt to push the incorrect type. I know it is not the desired practice, however, I included a second change in this build. This also supports a 'string' property in the C# class. The requirement to use the 'string' is that the PLC must be set to its default max (254) length.

Jason-Jelks:
I did not mean to "Close" this.

PR #394, opened 2021-05-31
Add synchronization

mycroes:


mycroes:
@scamille would you take a look at this?

scamille:
About the possibility of ensuring synchronization during open:
I think if you add a overloaded Enqueue Function accepting Func<Task> instead of Func<Task<T>> you could just use that in OpenAsync and guard the whole setup behind it. Which would be not that hard, or am I missing something else?

scamille (approved):
Looks good.

I am not familiar with the Inlocked.Exchange of Tasks, but I just assume this does what it is supposed to do ;)

mycroes:
> About the possibility of ensuring synchronization during open:
> I think if you add a overloaded Enqueue Function accepting Func instead of Func<Task> you could just use that in OpenAsync and guard the whole setup behind it. Which would be not that hard, or am I missing something else?

The issue is that `OpenAsync` relies on the `RequestTpdu` / `RequestTsdu` methods, so it would wait indefinitely. Not something I can't work around though, will try to get that in there today.

> I am not familiar with the Inlocked.Exchange of Tasks, but I just assume this does what it is supposed to do ;)

`Interlocked.Exchange` will just do a reliable exchange of values, the fact that I'm using it with `Task` shouldn't matter much. All unit tests still pass, I'll see if I can add some tests that try to enforce concurrency as well.

Thanks a lot for your review!

scamille:
Oh right, well I guess then it is either copy-paste time or you create some private versions of those request functions which work without posting their work to the queue.

mycroes:
Well `RequestTpdu` and `RequestTsdu` are already private, but `RequestTsdu` is actually only used by `OpenAsync` anyway... I'll try to turn it into something nice.

Issue #332, opened 2020-09-24
Connect S7-200 TPKT header is incomplete / invalid. Received Bytes: 0 expected: 4


hongphucbk:
I try connect to S7-200 using ethernet module CP 243-1 but has error.

![image](https://user-images.githubusercontent.com/35132643/94128901-e3d8dd00-fe84-11ea-8814-993cd6dc29a5.png)

![image](https://user-images.githubusercontent.com/35132643/94128965-fa7f3400-fe84-11ea-96e7-74701357f877.png)

I think default TSAP in S7-200 is difference with S7.net 
![image](https://user-images.githubusercontent.com/35132643/94129108-2a2e3c00-fe85-11ea-9466-9b53ad387542.png)

Please give me advise? Thank you

FalcoGoodbody:
please see #322 
This problem should be solved in Version 0.7.0.
Can please update to the newest one and retest?

mycroes:
Late to the show, but yes, I think your assesment is correct. Any chance you can test the code from #395?

ooowuyaooo:
> Late to the show, but yes, I think your assesment is correct. Any chance you can test the code from #395?

Hi bro 
you means here the lcoal and remote TSAP should be the same as 10.01 ? if set other values it can not be connected?

mycroes:
Hi @ooowuyaooo,

The screenshot from @hongphucbk shows the TSAP's are 10.01 and 10.00, if I understand the image correctly the PLC side is 10.01 and local 10.00. These values are quite essential in the S7 protocol, if they're incorrect the PLC will just close the connection (as evidenced by the exception). If the TSAP's from the screenshot are correct, the remote TSAP should be 10.01 (or `0x10`, `0x01`) instead of the previous value. I don't have such a PLC myself, so I can't test it myself either...

Regards,
Michael

ooowuyaooo:
Hi @mycroes 
First thank you for you codes it really helps me a lot.
yes i mean that TSAP values can be setted at will such as (local 12.01 remote 12.02), so are there availble to add this TSAP to the siemens s7-200 connection string ? like   Plc s7200plc = new Plc(CpuType.S7200, "10.136.176.3", 0, 2,"local tsap","remote tsap");
now everytime with the different TSAP i have to make lots of the dll file...
thanks bro 
regards
edis

mycroes:
@ooowuyaooo Yes, I'll add an option to specify TSAP's.

[...]

PR #396, opened 2021-06-03
Refactor TSAP related code

mycroes:


scamille (approved):

mycroes:
Still work in progress, thanks for the early review though!

scamille:
draft \*cough\* ;)

So far this is still quite a simple refactoring, but makes everything much more organized. I like the new version of the ```InvalidRackOrSlot``` function. There is still minor duplication of the min and max value, but that is ok since they are at least on the same line now.

mycroes:
I guess it's ready now though, asked some people in issues to test this, not sure if I'm gonna wait until they report back though, seeing as that might take months....

scamille:
As far as I understand those issues, the problem with the S7 200 connection already exists with the current version. So trying out the new value for the local tsap can't really make it worse, and if #332 is correct it should be resolved.

So probably best to just merge the refactoring, and even if the S7 200 connection still does not work it can always be given another try later.

Of course it would be nice to get a confirmation about it if you want to do a new release based on the changes.

mycroes:
Thanks again, merging as-is.

PR #397, opened 2021-06-06
Fix ReadMultipleVars alignment bug

### LINKED ISSUE #198: Reading multiple vars and a string with odd length fails
Hi 
When I read multiple vars and one of the vars is a StringEX with an odd legth, I get an exception: 
 PLC communication failed with error WrongNumberReceivedBytes

But this only occurs if the string is not the last item to read.  

I found the problem in class Plc (file PlcHelpers.cs) in method **ParseDataIntoDataItems**
The offset caculation is only adjusted for vartype byte or bit. I added String and StringEX.  

                // Fill byte in response when bytecount is odd
                if (dataItem.Count % 2 != 0 && (dataItem.VarType == VarType.Byte || dataItem.VarType == VarType.Bit 
                                             || dataItem.VarType == VarType.String || dataItem.VarType == VarType.StringEx ))
                    offset++;

Jürgen

scamille:
This PR tries to fix #198

### DIFF
--- a/S7.Net.UnitTest/S7NetTestsAsync.cs
+++ b/S7.Net.UnitTest/S7NetTestsAsync.cs
@@ -949,6 +949,33 @@ namespace S7.Net.UnitTest
             // Depending on how tests run, this can also just succeed without getting cancelled at all. Do nothing in this case.
             Console.WriteLine("Task was not cancelled as expected.");
         }
+
+        /// <summary>
+        /// Write a large amount of data and test cancellation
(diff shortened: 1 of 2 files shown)


mycroes:
Well at least the test works 😉

scamille:
It is not like I do much test-driven development at work, so I need to compensate at home ;)

mycroes:
Same for me. Merging this, I guess the type check was bollocks anyway....

PR #401, opened 2021-06-07
Add unit test to ensure a connection has been properly closed

scamille:
If canceled during data transmission, #349 adds code to ensure that the connection automatically gets canceled.

Add a test to ensure this works.
Also, add an inverted version to ensure that cancellation before sending data does not result in cancellation.

### DIFF
--- a/S7.Net.UnitTest/ConnectionCloseTest.cs
+++ b/S7.Net.UnitTest/ConnectionCloseTest.cs
@@ -79,7 +79,6 @@ namespace S7.Net.UnitTest
             var requestData = new byte[100]; // empty data, it does not matter what is in there
 
             // Set up access to private method and field
-            cancellationSource.CancelAfter(TimeSpan.FromMilliseconds(5));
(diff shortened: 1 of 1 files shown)


scamille:
Works on my machine :-/

mycroes:
I get the same error locally, pay attention that it only happens on net452, not the other targets.

mycroes:
Seems to me the issue is with `cancellationSource.CancelAfter(TimeSpan.FromMilliseconds(5))`, not sure why that's included in both tests (I think it shouldn't be in either...).

mycroes:
The fact that the tests now fail on a single platform is really good though. I still only have AppVeyor as required check, but I'm considering to make them all required.

scamille:
Right I thought I removed that line, that of course explains it, thanks.

It might sense to wait a bit more to see how reliable the new CI actions are. On the other I do know from the simulation craft/simc project that they are pretty reliable, since we have a lot more commits and longer tests over there, but not any .NET. 
Macos jobs can sometimes fail but it is still rare.

mycroes:
Thanks a lot @scamille!

Issue #414, opened 2021-09-30
ReadClass sbyte, uint, long, ulong aren't read


usingnamespaces:
I'm using the Method ReadClass.

For the datatypes 
DWord, have to use int as c# type instead of uint.
SInt, have to use byte as c# type instead of sbyte.
UDInt, have to use int as c# type instead of uint.

It's working meantime. I have to cast the properties back to the desired type.
Is it possible to fix this.

The PLC datatypes LWord, LInt, ULInt (all 64bit) aren't working at all if I declare them as long or ulong.

Jason-Jelks:
Which version of S7Net+ are you currently using?

usingnamespaces:
0.13.0

scamille:
Sbyte might not be supported, but a quick glance a the code suggests that uint should work fine.

What happens if you read a DWord or UDInt variable into a C# uint property?

usingnamespaces:
Structure of my datablock.
![2021-10-04_075930](https://user-images.githubusercontent.com/91665454/135801988-65cdce9a-d462-4d8f-8d34-3ca0fbddf57a.png)

Structure of the class.
![2021-10-04_080911](https://user-images.githubusercontent.com/91665454/135802027-049244a5-ac80-4443-8573-1a7b94c56bb8.png)


When variable 4 is read the actual value in the PLC is 3. The value read is 50331648.
For variable _11 the actual value in the PLC is 10. The value read is 167772160.

here the results
![2021-10-04_081405](https://user-images.githubusercontent.com/91665454/135802378-b5c5b409-a5e7-4a23-9ead-a501ae3112d6.png)

scamille:
Ok thanks for the data. I think I found the problem at https://github.com/S7NetPlus/s7netplus/blob/b475aee2e768f709e6b8f01084b176e53d43cd8c/S7.Net/Types/Class.cs#L169

Will try to create a fix for it.

Jason-Jelks:
For what its worth @scamille, I didn't respond to this thread yet, but I have almost completed the revisions for 64-bit values which I should be committing today. Up to you if you want to do it also.

mycroes:
I'm sorry guys, I'm a bit short on time (as usual), but I'll gladly accept a PR in similar style as other types.

PR #416, opened 2021-10-05
Jjelks - Resolves issues from #414 (added 64-bit value support)

Jason-Jelks:
Added support for LWord, LInt and ULInt PLC types.
Changed most byte[] passed parameters to Span<byte>, except in specific cases where performance and/or complexity outweighed benefits.
Class now supports all types except SInt and USInt. The SInt family is a single byte but gets allocated 2 bytes which conflicts with Int16 conversion process.
Added UnitTest for Class for 64-bit.


mycroes:
Hi Jason,

Thanks for your contribution. Contribution might actually be a huge understatement, because with a grand total of 70 files changed I'm guessing there's a lot more than just one contribution. Unfortunately that's also the problem with this PR.

As with commits I'd like to keep PR's scope limited to a single feature, but this PR contains so many changes I'm not even sure where to start reviewing. I took a peek through all the changed files and noticed there are quite some changes that are probably the result of some merges. I'm not sure if there are valid changes in there, but certainly there are some changes that shouldn't be merged (like a bunch of unimplemented unit tests).

My suggestion would be to create a new branch from the latest commit in this PR, do a mixed reset to S7NetPlus develop and stage and commit relevant changes only. You could repeat the procedure if there are other changes you want to contribute, like perhaps some of the changes to the unit tests. Another option would be that I do this myself, picking only the changes related to 64 bit sized data. I'd still be reviewing that code in that case and might change it as I see fit at that point, while I can also imagine you might actually want to be the one contributing these changes.

Let me know what would work best for you!

Regards,
Michael

Issue #422, opened 2021-12-26
OpenAsync with cancellation?


gfoidl:
Starting with .NET 5 `TcpClient` got an [overload that takes `CancellationToken`](https://apisof.net/catalog/89354e1a-28d2-06c4-e792-e9289384c4f2).

Should there be added a .NET 5 target to the project, so that the cancellation token can be flowed to https://github.com/S7NetPlus/s7netplus/blob/b475aee2e768f709e6b8f01084b176e53d43cd8c/S7.Net/PlcAsynchronous.cs#L28 ?

The issue https://github.com/S7NetPlus/s7netplus/issues/346 -- besided being closed -- could be fixed that way too, as a `CancellationTokenSource` can have a timeout.

---

I could submit a PR for this.

scamille:
Sure that might be helpful. While having code duplication and specialization based on the net version targeted isn't very nice and is generally avoided, I think in this case it is worth the trouble.

Just make sure to stick to the minimum required changes, without polluting the PR with unrelated things :-)

gfoidl:
> Just make sure to stick to the minimum required changes, without polluting the PR with unrelated things :-)

I'm not sure what you are talking about 🤣 
Of course I'll do here, otherwise this could be a mega-PR which isn't manageable...

mycroes:
> Any reason why this is still not merged? We would love to have this in production.

I guess just lack of time. Updating the PR as I'm writing this, if it builds I'll merge it.

dannyhaak:
Understandable and much appreciated. Thanks!

mycroes:
Should be available as 0.15.0 anytime now (NuGet is validating).

PR #423, opened 2021-12-26
Flowed cancellation token to TcpClient.ConnectAsync in .NET 5.0 target

gfoidl:
Fixes https://github.com/S7NetPlus/s7netplus/issues/422

---

I wanted to add a test similar to https://github.com/S7NetPlus/s7netplus/blob/b475aee2e768f709e6b8f01084b176e53d43cd8c/S7.Net.UnitTest/S7NetTestsAsync.cs#L39-L46 but then saw that due https://github.com/S7NetPlus/s7netplus/blob/b475aee2e768f709e6b8f01084b176e53d43cd8c/S7.Net.UnitTest/S7NetTestsSync.cs#L63-L68 the connection will always be established in the setup, so this test (and the one I'd like to add) are actually nops.

Personally I don't have a problem when there's no test, as the code just passes the cancellation token to the framework provided method, so there's not much to test anyway.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-image: Visual Studio 2019
+image: Visual Studio 2022
 configuration: Release
 install:
   - choco install gitversion.portable -y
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -1,4 +1,4 @@
-image: Visual Studio 2019
+image: Visual Studio 2022
```
gfoidl, line 1: This is to allow https://github.com/S7NetPlus/s7netplus/blob/5d3f01e59e61720b3dedd3342b433824b6287214/S7.Net/PlcAsynchronous.cs#L52

PR #435, opened 2022-06-16
Allow changing the default Encoding used in S7String

ismdiego:
The default Encoding.ASCII used in S7String to serialize and deserialize strings to/from byte array is not always valid in some situations where you want to use extended ASCII Code Pages, like the Windows 1252, or Iso-8859-1.

This simple PR allows user to change the Encoding system-wide, using an static property directly on the S7String class. I do not know if a central configuration already exists and this setting can be moved there. That would be nicer than this quick'n'dirty fix.

### DIFF
--- a/S7.Net/Types/S7String.cs
+++ b/S7.Net/Types/S7String.cs
@@ -9,10 +9,17 @@ namespace S7.Net.Types
     /// </summary>
     public static class S7String
     {
+        private static Encoding stringEncoding;
+
         /// <summary>
         /// The Encoding used when serializing and deserializing S7String (Encoding.ASCII by default)
         /// </summary>
-        public static Encoding StringEncoding { get; set; }
+        /// <exception cref="ArgumentNullException">StringEncoding must not be null</exception>
+        public static Encoding StringEncoding
+        {
+            get => stringEncoding ??= Encoding.ASCII;
+            set => stringEncoding = value ?? throw new ArgumentNullException(nameof(StringEncoding));
+        }
 
         /// <summary>
         /// Converts S7 bytes to a string
@@ -35,7 +42,6 @@ namespace S7.Net.Types
 
             try
             {
-                if (StringEncoding == null) StringEncoding = Encoding.ASCII;
                 return StringEncoding.GetString(bytes, 2, length);
             }
             catch (Exception e)
@@ -61,7 +67,6 @@ namespace S7.Net.Types
 
             if (reservedLength > 254) throw new ArgumentException($"The maximum string length supported is 254.");
 
-            if (StringEncoding == null) StringEncoding = Encoding.ASCII;
             var bytes = StringEncoding.GetBytes(value);
             if (bytes.Length > reservedLength) throw new ArgumentException($"The provided string length ({bytes.Length} is larger than the specified reserved length ({reservedLength}).");
 
@@ -70,14 +75,6 @@ namespace S7.Net.Types
             buffer[0] = (byte)reservedLength;
             buffer[1] = (byte)bytes.Length;
             return buffer;
-        }
-
-        /// <summary>
-        /// Initializes default static properties
-        /// </summary>
-        static S7String()
-        {
-            StringEncoding = Encoding.ASCII;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


S7.Net/Types/S7String.cs
```diff
@@ -9,6 +9,11 @@ namespace S7.Net.Types
     /// </summary>
     public static class S7String
     {
+        /// <summary>
+        /// The Encoding used when serializing and deserializing S7String (Encoding.ASCII by default)
+        /// </summary>
+        public static Encoding StringEncoding { get; set; }
```
gfoidl, line 15:
```suggestion
        public static Encoding StringEncoding { get; set; } = Encoding.ASCII;
```
* avoids the `null`-check in L38, and L64 (whole line can be deleted)
* doesn't need a static constructor (they aren't ideal, because they don't use `beforefieldinit` (in IL))

S7.Net/Types/S7String.cs
```diff
@@ -65,5 +71,13 @@ public static byte[] ToByteArray(string value, int reservedLength)
...
             return buffer;
         }
+
+        /// <summary>
+        /// Initializes default static properties
+        /// </summary>
+        static S7String()
+        {
+            StringEncoding = Encoding.ASCII;
+        }
```
gfoidl, line 81:
```suggestion
```
Not needed -- see comment above.

S7.Net/Types/S7String.cs
ismdiego: Yes, I agree. Initially, the checks in L38 and L64 were intended to avoid null exception if the user later sets the StringEncoding to null. But then, maybe it's ok to throw an exception in that case

S7.Net/Types/S7String.cs
ismdiego: Ok, nicer

S7.Net/Types/S7String.cs
gfoidl: Ah, good point. Maybe this should be a "regular property" where the setter has the null-check and if `null` throw an ArgumentNullException?
This would be the "most correct"* implementation.

\* except concerns about race-conditions, but assuming this static property will be set once at startup this shouldn't be a problem at all.

ismdiego: Excellent! I will change that now

S7.Net/Types/S7String.cs
```diff
@@ -8,11 +8,18 @@ namespace S7.Net.Types
...
+
+        /// <summary>
+        /// The Encoding used when serializing and deserializing S7String (Encoding.ASCII by default)
+        /// </summary>
+        /// <exception cref="ArgumentNullException">StringEncoding must not be null</exception>
+        public static Encoding StringEncoding
+        {
+            get => stringEncoding ??= Encoding.ASCII;
+            set => stringEncoding = value ?? throw new ArgumentNullException(nameof(StringEncoding));
+        }
```
gfoidl, line 22:
```suggestion
        private static Encoding stringEncoding = Encoding.ASCII;

        /// <summary>
        /// The Encoding used when serializing and deserializing S7String (Encoding.ASCII by default)
        /// </summary>
        /// <exception cref="ArgumentNullException">StringEncoding must not be null</exception>
        public static Encoding StringEncoding
        {
            get => stringEncoding;
            set => stringEncoding = value ?? throw new ArgumentNullException(nameof(StringEncoding));
        }
```
Can't be `null` now* as the setter validates that.
Otherwise LGTM.

\* except with some unsafe hacks

ismdiego: Yes, you are right. I am used to lazy initialization but this time is useless. I will change this right now. Thanks again 👍

ismdiego:
Last time I did a PR the appveyor part also failed. Should I take anything more into consideration when creating the PR? I think it's my fault...

gfoidl:
> I think it's my fault...

It's not your fault, and you can't do anything here. The failure is
```
GitVersion.BugException: GitVersion has a bug, your HEAD has moved after repo normalisation.
```

scamille (approved):

mycroes:
Thx guys. Yes GitVersion has some issues, might look into that at some point...

mycroes:
Released as 0.14.0, thanks again!

PR #437, opened 2022-06-20
Fix bytes number approximation

lucapolesel:
In case of a structure containing 9 bool type objects, the previous direct cast was rounding the resulting `numBytes` of `1.125` to `1` which is wrong because the last boolean is on the first bit of the second byte.

### DIFF
--- a/S7.Net/Types/Struct.cs
+++ b/S7.Net/Types/Struct.cs
@@ -77,7 +77,7 @@ namespace S7.Net.Types
                         break;
                 }
             }
-            return (int)numBytes;
+            return (int)Math.Ceiling(numBytes);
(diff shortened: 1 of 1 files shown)


scamille:
Thanks for your contribution.

Would it be possible for you to write a quick unit test for this? For example in here: https://github.com/S7NetPlus/s7netplus/blob/develop/S7.Net.UnitTest/TypeTests/ClassTests.cs

scamille (approved):

lucapolesel:
@scamille Sure, but I would like to keep the PR open for the time being until I find a better method to fix this problem (there are some more cases that I have to test).

PR #462, opened 2023-01-20
TimeoutException management in TPKT.ReadAsync

CSTTrevisiolPaolo:
TPKTInvalidException does't mean TimeoutException ... this manage real TimeoutException.

### DIFF
--- a/S7.Net/TPKT.cs
+++ b/S7.Net/TPKT.cs
@@ -33,7 +33,12 @@ namespace S7.Net
         public static async Task<TPKT> ReadAsync(Stream stream, CancellationToken cancellationToken)
         {
             var buf = new byte[4];
-            int len = await stream.ReadExactAsync(buf, 0, 4, cancellationToken).ConfigureAwait(false);
+
+            Task<int> ansTask = stream.ReadExactAsync(buf, 0, 4, cancellationToken);
+            ansTask.Wait(stream.ReadTimeout);
+            if (!ansTask.IsCompleted) throw new TimeoutException();
+
+            int len = ansTask.Result;
             if (len < 4) throw new TPKTInvalidException("TPKT is incomplete / invalid");
 
             var version = buf[0];
@@ -41,7 +46,12 @@ namespace S7.Net
             var length = buf[2] * 256 + buf[3]; //BigEndian
 
             var data = new byte[length - 4];
-            len = await stream.ReadExactAsync(data, 0, data.Length, cancellationToken).ConfigureAwait(false);
+
+            ansTask = stream.ReadExactAsync(data, 0, data.Length, cancellationToken);
+            ansTask.Wait(stream.ReadTimeout);
+            if (!ansTask.IsCompleted) throw new TimeoutException();
+
+            len = ansTask.Result;
             if (len < data.Length)
                 throw new TPKTInvalidException("TPKT payload incomplete / invalid");
 


S7.Net/TPKT.cs
```diff
@@ -33,15 +33,25 @@ private TPKT(byte version, byte reserved1, int length, byte[] data)
         public static async Task<TPKT> ReadAsync(Stream stream, CancellationToken cancellationToken)
         {
             var buf = new byte[4];
-            int len = await stream.ReadExactAsync(buf, 0, 4, cancellationToken).ConfigureAwait(false);
+
+            Task<int> ansTask = stream.ReadExactAsync(buf, 0, 4, cancellationToken);
+            ansTask.Wait(stream.ReadTimeout);
```
mycroes, line 38: This blocks the thread vs. the previous async operation.

S7.Net/TPKT.cs
```diff
@@ -33,15 +33,25 @@ private TPKT(byte version, byte reserved1, int length, byte[] data)
...
 
             var version = buf[0];
             var reserved1 = buf[1];
             var length = buf[2] * 256 + buf[3]; //BigEndian
 
             var data = new byte[length - 4];
-            len = await stream.ReadExactAsync(data, 0, data.Length, cancellationToken).ConfigureAwait(false);
+
+            ansTask = stream.ReadExactAsync(data, 0, data.Length, cancellationToken);
+            ansTask.Wait(stream.ReadTimeout);
```
mycroes, line 51: As above

mycroes (changes requested):
This changes async code into blocking code. I'm open to accept an implementation that doesn't block though.

S7.Net/TPKT.cs
CSTTrevisiolPaolo: ansTask.Wait(stream.ReadTimeout);  --> ansTask.Wait(cancellationToken);
maybe make it async code ...

S7.Net/TPKT.cs
CSTTrevisiolPaolo: ansTask.Wait(stream.ReadTimeout);  --> ansTask.Wait(cancellationToken);
maybe make it async code ...

scamille:
I'd recommend reading https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/cancel-async-tasks-after-a-period-of-time and https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.createlinkedtokensource?view=net-7.0

I'd also

mycroes:
@CSTTrevisiolPaolo please follow the advice from @scamille. The current code is still blocking the thread. It's actually now waiting on either the task or the passed in `cancellationToken`, while if cancellation was requested it will now throw a `TimeoutException` instead of properly handling the cancellation instead.

I'm not sure if you're aware, but you can use a `CancellationToken` with timeout to achieve what you're trying to do here, except that you won't get a `TimeoutException`. If that is what you need though, you could catch the `OperationCancelledException` and throw a `TimeoutException` with the `OperationCancelledException` as `innerException` argument.

I will reconsider this PR if it's still properly async and doesn't break other behavior.

CSTTrevisiolPaolo:
@mycroes , in last days I use this library for communication in production linking with a S7-1200 CPU and S7-1500 CPU.
I've spent much time on understanding of "TPKTInvalidException" customer get ... in debug i found Exception during EstablishConnection function; no problem in RequestConnection call ... but assert during SetupConnection call ...
So, first calling in ReadAsync was ok, second call in ReadAsync cause "TPKTInvalidException" ... but really i found a timeout in stream call, and get timeout exeption error ... so I understand there was an error on setting IP Address .
Managing TimeoutException help me very much.

PR #474, opened 2023-04-17
Modify rack/slot data type to byte

JIIIIL:


mycroes:
LGTM. I think GitVersion is having issues, will have to look into that before I can merge this...

gfoidl:
This is a breaking change (change in public contract).
What's the benefit of having `byte` instead of `short` here? Using only 1 bytes instead of 2 won't have any real effect on this type.
If something is changed I'd do `Int16` -> `short`.

mycroes:
One benefit is that you can't use negative values. I'm not sure why this ever was declared as short, because the allowed values are constrained between 0 and 15 either way...

I'm not necessarily opposed to this (breaking) change.

PR #483, opened 2023-05-30
Use System.Memory for < .NET 5 and avoid (some) unnecessary allocations

gfoidl:
Unfortunately https://github.com/S7NetPlus/s7netplus/pull/482 got merged too fast, so I couldn't have a look and provide my feedback over there, thus here's a PR with my suggestions 😉

* For targets before .NET 5 the _System.Memory_ package is used, which provides Span, and Memory for these targets
  So the code can be simplified a bit by re-using parts of the code

* internal `MemoryStreamExtension.WriteBytes` renamed to `MemoryStreamExtension.Write` so that the usage as extension solely becomes `stream.Write`, so for .NET 5 (when built with VS 2022 as CI does ?!) the C# compiler can avoid some allocations by refering to assembly's static data segment, namely for patterns like `package.Write(new byte[] { 0x1, 0x2, ... }`. See [sharplab example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGADAAQY4B0AkgPIDchhAxAHYCuANpwIbCcxSAEwCWAZz4CmRAMzkspAMKkA3oVIbSABygiAbjwAugiknIpSAIXYjOQgBIweQmFAAKPMAGseAcxgAFACyMAC20ACeAMqGsDyhpGKxTqEANKQirIak8RDsWaQAvKQ4AJTqmmoEmjWJyfGUAOq6xgGsMADupMARxgDaWAC6qqQy6WQAvqWM1bUaSXGhTS2B7V09/QCcwyqkWOlYKMTpOFgAHOkArMcl43ekN5PTFXMLKcsirWvdvTB9MjsHvcLFMZnN5vUls1PjBLL8AgENjBSrl8oZnrMNBNCBMgA) for this (the `PrivateImplementationDetails` refers to static data, no further allocation))

### DIFF
--- a/S7.Net/PLCHelpers.cs
+++ b/S7.Net/PLCHelpers.cs
@@ -1,7 +1,6 @@
 ﻿using S7.Net.Helper;
 using S7.Net.Protocol.S7;
 using S7.Net.Types;
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using DateTime = S7.Net.Types.DateTime;
@@ -18,13 +17,13 @@ namespace S7.Net
         private static void BuildHeaderPackage(System.IO.MemoryStream stream, int amount = 1)
         {
             //header size = 19 bytes
-            stream.WriteByteArray(new byte[] { 0x03, 0x00 });
+            stream.Write(new byte[] { 0x03, 0x00 });
             //complete package size
-            stream.WriteByteArray(Types.Int.ToByteArray((short)(19 + (12 * amount))));
-            stream.WriteByteArray(new byte[] { 0x02, 0xf0, 0x80, 0x32, 0x01, 0x00, 0x00, 0x00, 0x00 });
+            stream.Write(Int.ToByteArray((short)(19 + (12 * amount))));
+            stream.Write(new byte[] { 0x02, 0xf0, 0x80, 0x32, 0x01, 0x00, 0x00, 0x00, 0x00 });
             //data part size
-            stream.WriteByteArray(Types.Word.ToByteArray((ushort)(2 + (amount * 12))));
-            stream.WriteByteArray(new byte[] { 0x00, 0x00, 0x04 });
+            stream.Write(Word.ToByteArray((ushort)(2 + (amount * 12))));
+            stream.Write(new byte[] { 0x00, 0x00, 0x04 });
             //amount of requests
             stream.WriteByte((byte)amount);
         }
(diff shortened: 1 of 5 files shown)


S7.Net/PLCHelpers.cs
```diff
@@ -18,13 +17,13 @@ public partial class Plc
         private static void BuildHeaderPackage(System.IO.MemoryStream stream, int amount = 1)
         {
             //header size = 19 bytes
-            stream.WriteByteArray(new byte[] { 0x03, 0x00 });
+            stream.Write(new byte[] { 0x03, 0x00 });
             //complete package size
-            stream.WriteByteArray(Types.Int.ToByteArray((short)(19 + (12 * amount))));
-            stream.WriteByteArray(new byte[] { 0x02, 0xf0, 0x80, 0x32, 0x01, 0x00, 0x00, 0x00, 0x00 });
+            stream.Write(Int.ToByteArray((short)(19 + (12 * amount))));
+            stream.Write(new byte[] { 0x02, 0xf0, 0x80, 0x32, 0x01, 0x00, 0x00, 0x00, 0x00 });
```
gfoidl, line 23: These byte-array allocations are avoided for newer targets.
Could be done for older targets too (by using a static readonly field), but as this code was so before I don't think we should optimize for these targets -- and for the newer ones we get this optimization for free.

S7.Net/PlcSynchronous.cs
```diff
@@ -347,43 +321,18 @@ public void WriteClass(object classValue, int db, int startByteAdr = 0)
...
         private void ReadBytesWithSingleRequest(DataType dataType, int db, int startByteAdr, Span<byte> buffer)
         {
             try
             {
                 // first create the header
-                int packageSize = 19 + 12; // 19 header + 12 for 1 request
-                var package = new System.IO.MemoryStream(packageSize);
+                const int packageSize = 19 + 12; // 19 header + 12 for 1 request
+                var dataToSend = new byte[packageSize];
+                var package = new MemoryStream(dataToSend);
```
gfoidl, line 331: Instead of a `ToArray` on the MS we can just give it our buffer and use that buffer later on.
Avoids allocations (of the MS buffer).

mycroes:
Hi @gfoidl,

Sorry for acting so quickly 😆. I was considering some of the reductions you applied here, however I was very much satisfied with the fact someone else did most of the work on this. I fully agree on adding the System.Memory dependency though and making these improvements. Again your keen eye spotted some nice optimizations (the `Write` method naming).

I have to look into the build environment, not sure what platform we use on AppVeyor, which is what's used to deploy the package as well. I recently did my first GitHub actions workflow to publish to NuGet ([VisconFactoryIntelligence/AdsClient/blob/main/.github/workflows/dotnet.yml](https://github.com/VisconFactoryIntelligence/AdsClient/blob/main/.github/workflows/dotnet.yml), which also isn't actually based on VS2022, but I guess this works out with dotnet as well).

Anyway, thanks for your time and feedback, will merge and release this as well.

PR #490, opened 2023-07-19
Added support for serializing TimeSpan

bonk-dev:
TimeSpan can now be serialized to S7 Time (IEC), which is 4 bytes long.
The serialization works in structs too.
Tested on S7-1200.


mycroes (approved):

mycroes:
Thanks @bonk-dev, this will be included in the next release.

PR #499, opened 2023-08-14
Return expected length substring rather than throw an exception when reading strings.

Himmelt:
Sometimes users just want to read a part of the string, but if your count parameter is less than the string's length, it will cause an exception.


Himmelt:
Looks like I messed up? One pr with two functions change. I am not clear about the process of pull requests .....

mycroes:
Hi @Himmelt,

Just noticed this PR. Don't worry about making mistakes, we're all human! Great to see you chiming in and contributing code. I do suggest confirming if we'd be willing to accept a feature before spending a lot of time on it, but feel free to add as many PR's as you like. Also, don't be put off when something gets rejected, that's nothing personal and I always try to give a proper explanation why if I'm rejecting anything.

If you need any advice or help with Git / GitHub, feel free to ask, at the very least I can point you to the right help online or offer advice from personal experience.

Last but not least: Happy coding!

PR #501, opened 2023-08-15
Add SetBit method to modify one bit of a byte

Himmelt:
Add SetBit method to modify one bit of a byte

### DIFF
--- a/S7.Net/Conversion.cs
+++ b/S7.Net/Conversion.cs
@@ -155,9 +155,9 @@ namespace S7.Net
         /// Helper to set a bit value to the given byte at the bit index.
         /// Example: byte data = (byte)0x00; data.SetBit(4, true);// data -> 0x10
         /// </summary>
-        /// <param name="data">byte data to be modified</param>
-        /// <param name="index">bit index</param>
-        /// <param name="value">bool value to set</param>
+        /// <param name="data">The data to be modified.</param>
+        /// <param name="index">The zero-based index of the bit to set.</param>
+        /// <param name="value">The Boolean value to assign to the bit.</param>
         public static void SetBit(this ref byte data, int index, bool value)
         {
             if ((uint)index > 7)
--- a/S7.Net/Conversion.cs
+++ b/S7.Net/Conversion.cs
@@ -166,33 +166,13 @@ namespace S7.Net
 
             if (value)
             {
-                switch (bitPosition)
-                {
-                    case 0: data |= 0x01; return;
-                    case 1: data |= 0x02; return;
-                    case 2: data |= 0x04; return;
-                    case 3: data |= 0x08; return;
-                    case 4: data |= 0x10; return;
-                    case 5: data |= 0x20; return;
-                    case 6: data |= 0x40; return;
-                    case 7: data |= 0x80; return;
-                    default: return;
-                }
+                byte mask = (byte)(1 << bitPosition);
+                data |= mask;
             }
             else
             {
-                switch (bitPosition)
-                {
-                    case 0: data &= 0xFE; return;
-                    case 1: data &= 0xFD; return;
-                    case 2: data &= 0xFB; return;
-                    case 3: data &= 0xF7; return;
-                    case 4: data &= 0xEF; return;
-                    case 5: data &= 0xDF; return;
-                    case 6: data &= 0xBF; return;
-                    case 7: data &= 0x7F; return;
-                    default: return;
-                }
+                byte mask = (byte)~(1 << bitPosition);
+                data &= mask;
             }
         }
 
--- a/S7.Net/Conversion.cs
+++ b/S7.Net/Conversion.cs
@@ -147,7 +147,7 @@ namespace S7.Net
         ///   </code>
         /// </example>
         /// </summary>
-        /// <param name="data">The byte data to get from.</param>
+        /// <param name="data">The data to get from.</param>
         /// <param name="index">The zero-based index of the bit to get.</param>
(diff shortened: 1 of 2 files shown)


S7.Net/Conversion.cs
```diff
@@ -151,6 +151,51 @@ public static bool SelectBit(this byte data, int bitPosition)
...
 
+        /// <summary>
+        /// Helper to set a bit value to the given byte at the bit index.
+        /// </summary>
+        /// <param name="data"></param>
+        /// <param name="bitPosition"></param>
+        /// <param name="value"></param>
+        public static void SetBit(this ref byte data, int bitPosition, bool value)
+        {
+            if (bitPosition < 0 || bitPosition > 7)
```
gfoidl, line 162:
```suggestion
            if ((uint)bitPosition > 7)
```
With the `uint` cast any negative numbers will be interpreted as > int.MaxValue, thus > 7 and you can save one comparison / branch in the machine code.

S7.Net/Conversion.cs
```diff
@@ -151,6 +151,51 @@ public static bool SelectBit(this byte data, int bitPosition)
...
+                    case 1: data &= 0xFD; return;
+                    case 2: data &= 0xFB; return;
+                    case 3: data &= 0xF7; return;
+                    case 4: data &= 0xEF; return;
+                    case 5: data &= 0xDF; return;
+                    case 6: data &= 0xBF; return;
+                    case 7: data &= 0x7F; return;
+                    default: return;
+                }
+            }
```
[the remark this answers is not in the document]
gfoidl: Thanks for the translation 😄 

Just as info:
CPUs are very good at scalar operations (shift, negating, etc). Perf-wise there's only a register dependency (waiting for the previous result to be ready), but that's tiny to not really measurable.
On the other side CPUs are quite bad on branches if they are mis-predicted, as that causes a pipeline-stall which takes a lot of cpu-cycles. For the switch over bitPosition I don't think the cpu's branch predictor can predict the correct target, as the input (bitPosition) will very likely be random, thus a lot of pipeline-stalls will occur.

So for these kind of tasks scalar operations that are branchless are pretty much always prefereable.
Besides that: the code size is smaller.

S7.Net/Conversion.cs
```diff
@@ -151,6 +151,31 @@ public static bool SelectBit(this byte data, int bitPosition)
             return (result != 0);
         }
 
+        /// <summary>
+        /// Helper to set a bit value to the given byte at the bit index.
+        /// </summary>
+        /// <param name="data"></param>
```
mycroes, line 157: Add information to the `param` tags, currently they're all empty.

S7.Net/Conversion.cs
```diff
@@ -151,6 +151,31 @@ public static bool SelectBit(this byte data, int bitPosition)
...
+        /// <param name="bitPosition"></param>
+        /// <param name="value"></param>
+        public static void SetBit(this ref byte data, int bitPosition, bool value)
```
mycroes, line 160:
```suggestion
        public static void SetBit(this ref byte data, int index, bool value)
```

`bitPosition` is described as `index` in the method summary, I believe `index` is also what the exact name of the parameter should be.

mycroes (changes requested):
I'm open to accepting this, needs some minor changes.

S7.Net/Conversion.cs
Himmelt: 😂I was wanted to add these infos and use `index` instead of `bitPosition`, but when I saw `SelectBit` method and it's summary, to keep the same style and format, I wrote that codes.

I have updated with your suggested changes and modified the `SelectBit` method to keep the same style.

S7.Net/Conversion.cs
```diff
@@ -138,19 +138,45 @@ public static string ValToBinString(this object value)
...
-            int mask = 1 << bitPosition;
+            int mask = 1 << index;
             int result = data & mask;
 
             return (result != 0);
         }
 
+        /// <summary>
+        /// Helper to set a bit value to the given byte at the bit index.
+        /// Example: byte data = (byte)0x00; data.SetBit(4, true);// data -> 0x10
```
mycroes, line 156: The summary should just be that, the summary. The example can be moved into a separate block:

```csharp
  <example>
      Set the bit at index 4:
      <code>
        byte data = 0;
        data.SetBit(4, true);
      </code>
  </example>
```

S7.Net/Conversion.cs
```diff
@@ -138,19 +138,45 @@ public static string ValToBinString(this object value)
...
+        /// </summary>
+        /// <param name="data">byte data to be modified</param>
+        /// <param name="index">bit index</param>
```
mycroes, line 159:
```suggestion
        /// <param name="index">The zero-based index of the bit to set.</param>
```

S7.Net/Conversion.cs
```diff
@@ -138,19 +138,45 @@ public static string ValToBinString(this object value)
...
+        /// <param name="value">bool value to set</param>
```
mycroes, line 160:
```suggestion
        /// <param name="value">The Boolean value to assign to the bit.</param>
```

mycroes (changes requested):
Since you're on track, I thought let's make this perfect 🚀 Will you apply the suggestions to the SelectBit xmldoc as well?

mycroes (approved):

mycroes:
I'm not 100% sure examples should go in the summary, but I'll take another look at that after merging. Thanks for your contribution @Himmelt, I hope you enjoyed working on it!

Issue #505, opened 2023-08-22
"DATE" data type


TheGEZ:
Hi All !

Tis is my first post so first of all I want to thank everyone is supporting the awesome project, thank you very much !

As I said I'm new to S7Net+ and I have a question, may be stupid but ...
I'm developing a datalogging application that collects some data from an S7-300 PLC using the "ReadMultipleVars" function that has the amazing feature to automatically converts values in the corriponding C# data type, and I've noticed that seems that the "old" type DATE (16Bit IEC 61131-3  "number of days from 1-1-1990") is not directly supported, i I request the data using the "VarType.DateTime" type I get an error.
To be clear, this is not a big deal sice the data ca be read by using the "VarType.Word" type en then convert it to the C# DateTime using C# DateTime methods, but i think that would be very handy if the fuction could hande the type by itself.
So the question is, is the "DATE" type effectively unsupported or I missed something ?

Thank you again !

bonk-dev:
If the PR gets merged, you can clone this repo, build the solution, add the reference to your 
project (while you wait for the official release) and use the new `VarType.Date` like that:
`var dataItems = new List<DataItem>()
{
    new DataItem()
    {
        VarType = VarType.Date,
        Count = 1,
        DataType = DataType.DataBlock,
        DB = 1
    }
};
var result = await plc.ReadMultipleVarsAsync(dataItems);`

PR #506, opened 2023-08-22
Add support for reading/writing the legacy DATE (IEC date) datatype

bonk-dev:
Solves #505 

I did not bother with adding this type to struct serialization, because the DateTime type itself is not supported either.
I might create a separate PR and add both the DateTime and IEC date to struct later on.

### DIFF
--- a/S7.Net/Types/Date.cs
+++ b/S7.Net/Types/Date.cs
@@ -15,8 +15,14 @@ namespace S7.Net.Types
         
         /// <summary>
         /// Maximum allowed date for the IEC date type
+        /// <remarks>
+        /// Although the spec allows only a max date of 31-12-2168, the PLC IEC date goes up to 06-06-2169 (which is the actual
+        /// WORD max value - 65535) 
+        /// </remarks>
         /// </summary>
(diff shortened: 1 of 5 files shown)


S7.Net/Types/Date.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    /// <summary>
+    /// Contains the conversion methods to convert Words from S7 plc to C#.
+    /// </summary>
+    public static class Date
+    {
+        /// <summary>
+        /// Minimum allowed date for the IEC date type
+        /// </summary>
+        public static readonly System.DateTime IecMinDate = new(year: 1990, month: 01, day: 01);
```
gfoidl, line 14:
```suggestion
        public static System.DateTime IecMinDate { get; } = new(year: 1990, month: 01, day: 01);
```
To expose it as property instead of field?

Same below for `IecMaxDate`.

mycroes:
Hi @bonk-dev, sorry for my late response. Thankfully @gfoidl did a review quickly and I agree with his comments. Only thing that would've been nice is test-cases for the new conversion, but I'll accept this as-is. Thanks a lot for your contribution!

PR #508, opened 2023-09-12
add cpu status enum

LUJIAN2020:


S7.Net/PlcSynchronous.cs
```diff
@@ -520,12 +520,11 @@ public void WriteClock(System.DateTime value)
         /// Read the current status from the PLC. A value of 0x08 indicates the PLC is in run status, regardless of the PLC type.
         /// </summary>
         /// <returns>The current PLC status.</returns>
-        public byte ReadStatus()
+        public CpuStatus ReadStatus()
```
mycroes, line 523: You're missing a similar change for PlcAsynchronous.cs

PR #528, opened 2024-05-14
Fix reusing Plc after Close()

ltjax:
We got a `System.ObjectDisposedException` when trying to access a `Close()`d connection. This is because the cached `Plc._stream` was not reset when its parent `TcpClient` was reset.


gfoidl (approved):

ltjax:
Anything I can do to get this PR merged? Is it just stale because of the macOS check failing?

PR #534, opened 2024-07-17
Two byte alignment issue in the structure

fxyc87:


S7.Net/Types/Struct.cs
```diff
@@ -75,11 +76,15 @@ public static int GetStructSize(Type structType)
                         break;
                     default:
                         numBytes += GetStructSize(info.FieldType);
+                        if (count != infos.Count()) {
+                            if(numBytes % 2 != 0) 
+							    numBytes++;//word align
```
gfoidl, line 85:
```suggestion
                        if (count != infos.Count()) 
                        {
                            if (numBytes % 2 != 0) 
							    numBytes++;     //word align
```

S7.Net/Types/Struct.cs
```diff
@@ -238,6 +243,8 @@ public static int GetStructSize(Type structType)
                         Buffer.BlockCopy(bytes, (int)Math.Ceiling(numBytes), buffer, 0, buffer.Length);
                         info.SetValue(structValue, FromBytes(info.FieldType, buffer));
                         numBytes += buffer.Length;
+                        if(numBytes %2 != 0)        //word align
```
gfoidl, line 254:
```suggestion
                        if (numBytes %2 != 0)        //word align
```

gfoidl (commented):
Some style-nit.

Out of interest: did you run in any problem or was this found by inspecting the code?

S7.Net/Types/Struct.cs
```diff
@@ -9,12 +9,12 @@ namespace S7.Net.Types
     /// </summary>
     public static class Struct
     {
-        /// <summary>
-        /// Gets the size of the struct in bytes.
-        /// </summary>
-        /// <param name="structType">the type of the struct</param>
-        /// <returns>the number of bytes</returns>
-        public static int GetStructSize(Type structType)
+		/// <summary>
```
gfoidl, line 12: Nit: keep the spaces, so undo this change.

S7.Net/S7.Net.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
-    <TargetFrameworks>net452;net462;netstandard2.0;netstandard1.3;net5.0;net6.0;net7.0</TargetFrameworks>
```
gfoidl, line 4: I don't know the policy for supporting older targets, but the same could apply to .NET 5 too.

My feeling is that for PLCs quite a lot of older targets are around, because they run and an running system isn't touched (in that area). 

So maybe NS 1.3 should be kept too (except everything from these API are covered by net452 or net462 -- which I don't know as NS 1.3 is soo old 😉).

S7.Net/Types/Struct.cs
fxyc87: In a block of the PLC, two structures A and B are defined. Structure A occupies 3-byte space, and structure B also occupies 3-byte space. However, in reality, structure B starts from the fourth byte because the PLC is two-byte aligned internally. 

当在PLC一个块中，定义一个结构体A，B，其中A占用3个字节空间，B也占3个字节空间，但实际上B是从第4字节开始的，因为PLC内部是两字节对齐的。
将以上翻译为英文

S7.Net/Types/Struct.cs
fxyc87: When no user-defined structure is in a block of the PLC, there are no such problems. When a structure exists, there is a two-byte alignment issue.

当在PLC一个块中没有自定义结构体时，这些没有问题，当有结构体存在时，就有2字节对齐问题。

S7.Net/S7.Net.csproj
fxyc87: Sorry, this is not necessary. This happened when I was testing another code and the compilation failed. Later, I forgot to restore it.

S7.Net/Types/Struct.cs
fxyc87: The above code has been verified by the project and the test is normal.

gfoidl: My suggestion is about code style (see the spacing and newline).

PR #555, opened 2025-05-20
feat: Introduce TcpClientFactory

seilc1:
Allows injecting custom TcpClient for monitoring and logging

### DIFF
--- a/S7.Net/Tcp/ITcpClient.cs
+++ b/S7.Net/Tcp/ITcpClient.cs
@@ -1,0 +1,21 @@
+﻿using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace S7.Net.Tcp
+{
+    public interface ITcpClient
+    {
+        public int ReceiveTimeout { get; set; }
+
+        public int SendTimeout { get; set; }
+
+        public bool Connected { get; }
+
+        public void Close();
+
+        public Task ConnectAsync(string ip, int port, CancellationToken cancellationToken = default);
+
+        public NetworkStream GetStream();
+    }
+}
--- a/S7.Net/Tcp/TcpClientWrapper.cs
+++ b/S7.Net/Tcp/TcpClientWrapper.cs
@@ -1,0 +1,24 @@
+﻿using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace S7.Net.Tcp;
+
+internal class TcpClientWrapper : TcpClient, ITcpClient
+{
+    public async Task ConnectAsync(string ip, int port, CancellationToken cancellationToken)
+    {
+#if NET5_0_OR_GREATER
+        await base.ConnectAsync(ip, port, cancellationToken).ConfigureAwait(false);
+#else
+        await base.ConnectAsync(ip, port).ConfigureAwait(false);
+#endif
+    }
+
+    public void Close()
+    {
+#if NET20_OR_GREATER
+        base.Close();
+#endif
+    }
+}
--- a/S7.Net/Tcp/TcpClientWrapperFactory.cs
+++ b/S7.Net/Tcp/TcpClientWrapperFactory.cs
@@ -1,0 +1,9 @@
+﻿namespace S7.Net.Tcp;
+
+internal class TcpClientWrapperFactory : ITcpClientFactory
+{
+    public ITcpClient Create()
+    {
+        return new TcpClientWrapper();
+    }
+}
--- a/S7.Net.UnitTest/ConnectionCloseTest.cs
+++ b/S7.Net.UnitTest/ConnectionCloseTest.cs
@@ -1,11 +1,10 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 using System.IO;
-using System.Linq;
-using System.Net.Sockets;
 using System.Reflection;
 using System.Threading;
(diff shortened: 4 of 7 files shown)


ynioba:
这是来自”麦冬@广西机械工业研究院”的邮箱自动回复邮件。您好，您的邮件已收到，我查阅后尽快回复您。

S7.Net/Tcp/ITcpClient.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public int SendTimeout { get; set; }
+
+        public bool Connected { get; }
+
+        public void Close();
+
+        public Task ConnectAsync(string ip, int port, CancellationToken cancellationToken);
+
+        public Task ConnectAsync(string ip, int port);
```
gfoidl: Can the CT have a default-value (here just the empty one), so one overload can be avoided?
```suggestion
        public Task ConnectAsync(string ip, int port, CancellationToken cancellationToken = default);

```

S7.Net/Tcp/TcpClientWrapper.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace S7.Net.Tcp;
+
+public class TcpClientWrapper : TcpClient, ITcpClient
```
gfoidl:
```suggestion
internal class TcpClientWrapper : TcpClient, ITcpClient

```
Or is it meant to be visible outside of the lib too?

S7.Net/Tcp/TcpClientWrapper.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+
+namespace S7.Net.Tcp;
+
+public class TcpClientWrapper : TcpClient, ITcpClient
+{
+    public async Task ConnectAsync(string ip, int port, CancellationToken cancellationToken)
+    {
+#if NET5_0_OR_GREATER
+            await base.ConnectAsync(ip, port, cancellationToken).ConfigureAwait(false);
```
gfoidl:
```suggestion
        await base.ConnectAsync(ip, port, cancellationToken).ConfigureAwait(false);

```

S7.Net/Tcp/TcpClientWrapperFactory.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace S7.Net.Tcp;
+
+public class TcpClientWrapperFactory : ITcpClientFactory
```
gfoidl:
```suggestion
internal class TcpClientWrapperFactory : ITcpClientFactory

```
Same question here?

gfoidl (commented):
> custom TcpClient for monitoring and logging

Seems reasonable to me.

---

A different approach -- more or less a quite big rewrite -- would be incorporate logging, tracing, and metrics into the lib. But that's a lot of work, and older target frameworks won't profit from that change. So your approach seems to be the better way here.

seilc1:
> > custom TcpClient for monitoring and logging
> 
> Seems reasonable to me.
> 
> A different approach -- more or less a quite big rewrite -- would be incorporate logging, tracing, and metrics into the lib. But that's a lot of work, and older target frameworks won't profit from that change. So your approach seems to be the better way here.

Besides being a bigger rewrite it's also a question of what's needed. I'm not sure if "my" need for logging/metrics would fit a general need or implementation style. So I think it's better to just provide the "possibility".

S7.Net/Tcp/ITcpClient.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public int SendTimeout { get; set; }
+
+        public bool Connected { get; }
+
+        public void Close();
+
+        public Task ConnectAsync(string ip, int port, CancellationToken cancellationToken = default);
+
+        public Task ConnectAsync(string ip, int port);
```
gfoidl: W/ the CT = default, this overload can be removed.
```suggestion

```

gfoidl (approved):
LGTM -- modulo one comment.

I'd like to hear what the others think about this.

gfoidl:
Please don't force push a branch when a PR is open, that makes reviewing harder, as the last diff disappears.
Just push the normal commit to the branch.

When merging the PR, then (IMO ideally) a squash-merge is done to get the linear history in the main-branch again.

seilc1:
@gfoidl : any idea when your peers will react?

mycroes:
> @gfoidl : any idea when your peers will react?

Sorry for my late response, awfully busy lately... This seems to be fine, I need to fix the workflows though so GH can also verify it builds. I'll try to get this fixed ASAP.

seilc1:
> > @gfoidl : any idea when your peers will react?
> 
> Sorry for my late response, awfully busy lately... This seems to be fine, I need to fix the workflows though so GH can also verify it builds. I'll try to get this fixed ASAP.

Can I support somhow?

mycroes:
> > > @gfoidl : any idea when your peers will react?
> > 
> > 
> > Sorry for my late response, awfully busy lately... This seems to be fine, I need to fix the workflows though so GH can also verify it builds. I'll try to get this fixed ASAP.
> 
> Can I support somhow?

See PR #557. Unfortunately the unit tests don't run reliably, I'm not 100% sure of the cause, but it might be that the tests execute concurrently thus causing the failures. Feel free to chime in.

Issue #235, opened 2019-09-10
Bug in Struct.GetStructSize() ?


Equino3108:
I got a struct with size 16 Bytes and at the end, I also define 4 bool(bit) fields, which brings size up to 16.5 as GetStructSize adds 0.125 per bool field it finds.

At the end of GetStructSize, you return (int)numbytes which cuts off the .5 and makes the method return a size of 16.  After that it crashes at some point in Struct.ToBytes() - apparently because it tries to write bit values to byte 17 which isn't present i the bytes[] array.

I believe GetStructSize() should return (int)Math.Ceiling(numBytes) instead.

mycroes:
Sorry for leaving this without response for so long, but yes, you're absolutely right about this. I'm not sure if it should round to an even number of bytes as well, but I think it might. Marking this as a reminder, let me know if this has high priority for you.

scamille:
As far as I know a UDT also always starts at an alignment 2 bytes, or in other words at an even address.
The same is true for for data types with size 2 or more, at least for S7-1500. 

So assuming the C# struct tries to mimic the S7 one without the need for filler bytes, we should also handle that.

Probably does not happen that often since you PLC programmers don't like to waste bytes or even bits :-P

FalcoGoodbody:
Correct UDTs always starts at an alignment 2 bytes because S7 works word-oriented. Also variables within an datablock from type word, int, dint and real always start at even address.

scamille:
Can you explain what optimizing data blocks really does on the Plc?
Is it just rearranging things, so that serialization gets more complicated?
Can you still read the whole DB as once byte-array, assuming you know the optimized total size?

The problem, besides the effort to implement would definitely be figuring out how Siemens does this DB optimization. If there is some well-defined algorithm it would be still quite a huge undertaking; if not it also requires a ton of reverse-engineering.

And if optimized means you can no longer just read a raw amount of bytes from the data block, then there is probably no solution for it at all.

FalcoGoodbody:
Optimized Datablocks are very good evolution in the S7 plc world.
When using optimized data blocks the memory usage of the data block is managed by the system. There are no spaces between the variables and one can create structs in a way they want.

A good explanation gives the programming guideline from SIEMENS which can be found in englisch language [here](https://cache.industry.siemens.com/dl/files/040/90885040/att_970576/v1/81318674_Programming_guideline_DOC_v16_en.pdf) and is available in other languages too.

[...]

Issue #210, opened 2019-03-06
Wrong size returned by Struct.GetStructSize on some structs


rashandan:
In example:
`struct Bridge 
{
  public float float1; 
  public bool bool1;
  public bool bool2;
  public bool bool3;
  public bool bool4;
}`
size of abowe structure must be 5, but when reading by ReadStruct, it spawns exeption.
Inside this function, call to Struct.GetStructSize returns 4.
I believe there must be something like `return (int)Math.Ceiling(numBytes);` instead of `return (int)numBytes;`
This happens when struct ends by from 1 to 7 bool-s.

//sorry for my English

tsheyu:
I also find this problem before. Since in my program scenario, the struct contains struct. So the size must be Even -aligned. 
In your case  the size should be 6 bytes. U can open the data block to check the real size.
 
If it is confirmed, At the end of the function  "Struct.GetStructSize" add the code like
` numBytes = Math.Ceiling(numBytes);
                    if ((numBytes / 2 - Math.Floor(numBytes / 2.0)) > 0)
                        numBytes++;
                    break;`

mabc666:
the author round the size of struct it looks like a bug and hasn't been fixed  yet.
you can round up the size of struct 
`return (int)Math.Ceiling(numBytes); `
It can slove this problem

Issue #206, opened 2019-02-09
How to let function "ReadStruct" to support Array using Reflection


tsheyu:
Hi, Michael：
            in "S7.net\\Types\\Struct.cs" 
   ```
                    case "Double":
                        numBytes = Math.Ceiling(numBytes);
                        if ((numBytes / 2 - Math.Floor(numBytes / 2.0)) > 0)
                            numBytes++;
                        numBytes += 4;
                        break;
                    default:
                        numBytes += GetStructSize(info.FieldType);
                        break;
```
I think in the Default case also need a function
  ```
numBytes = Math.Ceiling(numBytes);
  if ((numBytes / 2 - Math.Floor(numBytes / 2.0)) > 0)
                        numBytes++;
```

Functions "FromBytes", "ToBytes", "GetStructSize" are affected. Please have a look~~~

PR #556, opened 2025-05-30
Struct.cs fix counting length to read

stefanak-michal:
Fixed count when there is amount of boolean values which doesn't count up to one byte.


stefanak-michal:
https://github.com/S7NetPlus/s7netplus/issues/559
https://github.com/S7NetPlus/s7netplus/issues/235
https://github.com/S7NetPlus/s7netplus/issues/210
https://github.com/S7NetPlus/s7netplus/issues/206

PR #557, opened 2025-06-02
Update GitHub actions

mycroes:


seilc1:
I think we had a similar issue in the recent past: the ports used are not that eagerly freed, so that if a fixed port is used, it can happen that it's still considered in use.

There are (atleast) two unreliable solutions:
1. increment the port each time -> increased potential to conflict with other applications
2. try to get a free port by using `new TcpListener(IPAddress.Loopback, 0)` -> can still be flaky in multi threading scenarios

what's the best way to provide you a fix suggestion?

seilc1:
```c#
    [TestClass]
    public partial class S7NetTests : IDisposable
    {
        #region Constants
        const int DB2 = 2;
        const int DB4 = 4;
        const short TestServerPortStart = 31122;
        const string TestServerIp = "127.0.0.1";
        #endregion

        #region Private fields
        Plc plc;
        private readonly short testServerPort;
        #endregion

        private static readonly object _lock = new();
        private static short currentTestServerPort = TestServerPortStart;

        #region Constructor
        /// <summary>
        /// Create a plc that will connect to localhost (Snap 7 server) and connect to it
        /// </summary>
        public S7NetTests()
        {
            lock (_lock)
            {
                currentTestServerPort++;
                testServerPort = currentTestServerPort;
            }

            plc = CreatePlc(testServerPort);

        }

        private static Plc CreatePlc(short port)
        {
            return new Plc(CpuType.S7300, TestServerIp, port, 0, 2);
        }

        [TestInitialize]
        public void Setup()
        {
            S7TestServer.Start(testServerPort);
            plc.Open();
        }
```

PR #562, opened 2026-02-11
Fixed a potential memory leak risk in `MemoryStreamExtension.Write`

Vincent-X-Zhang:
Move the function `ArrayPool<byte>.Shared.Return` to the finally block.

### DIFF
--- a/S7.Net/Helper/MemoryStreamExtension.cs
+++ b/S7.Net/Helper/MemoryStreamExtension.cs
@@ -28,10 +28,15 @@ namespace S7.Net.Helper
         {
             byte[] buffer = ArrayPool<byte>.Shared.Rent(value.Length);
 
-            value.CopyTo(buffer);
-            stream.Write(buffer, 0, value.Length);
-
-            ArrayPool<byte>.Shared.Return(buffer);
+            try
+            {
+                value.CopyTo(buffer);
+                stream.Write(buffer, 0, value.Length);
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(buffer);
+            }
         }
     }
 #endif


ynioba:
这是来自”麦冬@广西机械工业研究院”的邮箱自动回复邮件。您好，您的邮件已收到，我查阅后尽快回复您。

S7.Net/Helper/MemoryStreamExtension.cs
```diff
@@ -28,10 +28,15 @@ public static void Write(this MemoryStream stream, ReadOnlySpan<byte> value)
...
-
-            ArrayPool<byte>.Shared.Return(buffer);
+            try
+            {
+                value.CopyTo(buffer);
+                stream.Write(buffer, 0, value.Length);
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(buffer);
```
gfoidl, line 38: The try-finally is actually not what one wants here. See https://github.com/dotnet/runtime/discussions/48257#discussioncomment-14909499 and the following comments.

Thanks for trying to address a problem anyway!
But I'd recommend to close this PR due the reasons mentioned in the link above.

Vincent-X-Zhang: I also recommend introducing `System.IO.Pipelines` to replace `MemoryStream` for improved performance. And what do you think?

gfoidl: > what do you think?

I think the whole codebase is quite aged and could be put onto new feets. But this is a big effort.
Moving to Pipelines may be one step in that direction, and I would welcome such a change.

mycroes: In all honesty I'm not sure if it wouldn't be better to concentrate efforts on different libraries. Sally7 for instance is way more memory optimized than S7NetPlus (and used in production for over 10 years), but uses the same outdated protocol as S7NetPlus. OTOH there's new efforts like [thomas-v2/S7CommPlusDriver](https://github.com/thomas-v2/S7CommPlusDriver) that support newer protocols.

One of the benefits of S7NetPlus is that it allows reading classes and structs, but the implementation is quirky, has no automated tests (like most of S7NetPlus for that matter) and could generally be hugely improved. Regardless I'm not confident about the value of this part, I haven't had the need to read classes or structs personally and one could still just read byte arrays and do conversion afterwards.

If you'd still want to put in effort into S7NetPlus my suggestion would be to start creating test coverage and go from there.

Vincent-X-Zhang: I tried to fork a repository's code and make modifications. Here’s what I plan to do: 
1. Refactor the method body without changing the API signature.
2. Add more unit tests.
3. I will push the code once the tests are thorough and appropriate.

Welcome to wathc my repository and provide any good suggestions. Thx! @mycroes @gfoidl

